PR #3, opened 2025-01-19
feat: add progress notifications

PeterCardenas:
The current progress notifications are only fit for VSCode, but we can use the LSP $/progress method to notify for progress which will work for any LSP clients that use the method.


CppCXY:
thank  you

PR #21, opened 2025-02-04
add basic table field names autocompletion

### LINKED ISSUE #20: Completions for table declarations fields which types can be inferred
It'd nice to have completions for field names when declaring tables with types that can be inferred. For example.

```lua
---@class Bar: table
---@field buz number

---@param a Bar
local function foo(a)
end

foo({
    -- Completion for `buz` here...
})
```

georgiy-belyanin:
This patch adds basic autocompletion features for table field names when declaring a table.

Example.
```lua
---@class Bar: table
---@field buz number

---@param a Bar
local function foo(a)
end

foo({
    -- `buz` field is suggested here...
})
```

The implementation is very basic. Known limitations are the following.
* The field names are suggested only when the table is instantiated within the function arguments with defined types.
* Generics/inferred funkjction types aren't supported yet.

Closes #20

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 `NEW` support `code style check`, which powered by `emmyluacodestyle`
 
+`NEW` Basic table declaration field names autocompletion.
+
 # 0.4.3
 
 `FIX` Fix std resource loaded for cli tools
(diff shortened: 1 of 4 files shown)


georgiy-belyanin:
I've implemented only the very basic autocompletion but it's really handy for the functions defining kind of named `opts` table arguments.

PR #22, opened 2025-02-04
fix possible panic when calculating pow values

georgiy-belyanin:
This patch fixes possible LSP panic that occurs when LSP type inferer power calculations overflowed the `i64` value upper bound.

The fix is straightforward: replace the `pow()` call that may throw an exception with safe `checked_pow()`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 `NEW` Basic table declaration field names autocompletion.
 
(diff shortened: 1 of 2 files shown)


georgiy-belyanin:
Force-pushed due to a typo. Sorry.

This is a fix of a bug I came across when checking a real project. Probably, the panic happened due to `2^64` expressions in the project code.

PR #25, opened 2025-02-05
add function parameter types to completions

georgiy-belyanin:
This patch adds function parameter types to completion suggestions to make it easier for user to understand the parameter type. For instance.

```lua
---@param a number
local function test(a) end
```

Previous suggestion: `test Function (a)`.
New suggestion: `test Function (a: number)`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 # 0.4.5(unreleased)
 
+`NEW` Function completion suggestions now contain type parameters.
 
 # 0.4.4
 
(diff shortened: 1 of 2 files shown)


georgiy-belyanin:
This PR is kind of a suggestion. Feel free to close if you don't like it.

image

CppCXY:
Having observed many IDEs for various programming languages, they generally do not choose to display the types of function parameters on the completion item because they can be very lengthy. In fact, the more appropriate place to display the function signature is in the documentation. I am not sure why you do not have this. Additionally, I have also displayed the parameter types of functions in the signature help. I am uncertain whether the editor has implemented these features.

georgiy-belyanin:
Feels like NeoVim isn't able to display the type of the prameters.

image

Ok then. Closing the PR than. I'll work on displaying the parameter types as a part of the documentation.

PR #42, opened 2025-02-06
infer constant array tables as tuples

### LINKED ISSUE #40: Support inferring constant arrays and their indices
It'd nice to support inferring constant arrays and their indices types. Now, when defining a constant array the inferred type is the first field, which is imprecise.

Current behavior.
```lua
-- Inferred as table const, humanized like `table`
local a = { 'a', 1 }

-- Inferred as `'a'`
local b = a[2]
```

LSP may infer such constant array as a tuple or as a const table having keys `1` and `2`. Probable desired behavior.

```lua
-- Inferred as ['a', 1].
local a = { 'a', 1 }

-- Inferred as `1`
local b = a[2]
```

georgiy-belyanin:
This patch makes LSP infer constant array tables as tuples of elements.

Example.
```lua
-- Inferred `(1, 2)`
local a = {1, 2}

-- Inferred `("test",(3,4))`
local b = {'test', {3, 4}}
```

Thus, making preciser inference of indexing statements.

```lua
--- Inferred `4`
local c = b[2][1]
```

Closes #40

### DIFF
--- a/crates/emmylua_code_analysis/src/semantic/infer/infer_table.rs
+++ b/crates/emmylua_code_analysis/src/semantic/infer/infer_table.rs
@@ -1,12 +1,24 @@
 use emmylua_parser::{LuaAstNode, LuaTableExpr};
 
-use crate::db_index::{DbIndex, LuaType};
+use crate::db_index::{DbIndex, LuaType, LuaTupleType};
 
-use super::{InferResult, LuaInferConfig};
+use super::{InferResult, LuaInferConfig, infer_expr};
 
-pub fn infer_table_expr(_: &DbIndex, config: &LuaInferConfig, table: LuaTableExpr) -> InferResult {
-    Some(LuaType::TableConst(crate::InFiled {
-        file_id: config.get_file_id(),
-        value: table.get_range(),
-    }))
+pub fn infer_table_expr(db: &DbIndex, config: &mut LuaInferConfig, table: LuaTableExpr) -> InferResult {
+    if table.is_array() {
+        infer_table_array_expr(db, config, table)
+    } else {
+        Some(LuaType::TableConst(crate::InFiled {
+            file_id: config.get_file_id(),
+            value: table.get_range(),
+        }))
+    }
+}
+
+fn infer_table_array_expr(db: &DbIndex, config: &mut LuaInferConfig, table: LuaTableExpr) -> InferResult {
+    let field_types = table.get_fields()
+        .map(|field| infer_expr(db, config, field.get_value_expr()?))
+        .map(|field_type| field_type.unwrap_or(LuaType::Unknown))
+        .collect::<Vec<_>>();
+    Some(LuaTupleType::new(field_types).into())
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@
 
 `NEW` Compact luals env variable start with `$`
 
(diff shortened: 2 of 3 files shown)


crates/emmylua_code_analysis/src/semantic/infer/infer_table.rs
```diff
@@ -1,12 +1,24 @@
...
+    } else {
+        Some(LuaType::TableConst(crate::InFiled {
+            file_id: config.get_file_id(),
+            value: table.get_range(),
+        }))
+    }
+}
+
+fn infer_table_array_expr(db: &DbIndex, config: &mut LuaInferConfig, table: LuaTableExpr) -> InferResult {
+    let field_types = table.get_fields()
```
CppCXY, line 19: I think it should be inferred as a tuple only if every item is a constant expression; otherwise, it should be treated according to the first item of the array. Additionally, considering the number of items, if the array length exceeds 10 items, it should not be calculated as a tuple; it is likely a configuration table.

georgiy-belyanin: > I think it should be inferred as a tuple only if every item is a constant expression; otherwise, it should be treated according to the first item of the array.

Why do you think LSP should infer less precise type? I suppose this approach can only be applied if every element of a table is of a same type, since the user is still able to write `{<number>, <string>}` with variables and inferring a type based on the first element (or throw a warning on the second since it's `string`, not a `number`) won't work.

> Additionally, considering the number of items, if the array length exceeds 10 items, it should not be calculated as a tuple; it is likely a configuration table

You're right. I'll apply it.

CppCXY: Lua is not a statically typed language, and usually, there is not enough information to infer the best type of an expression. Additionally, Lua does not have a precise startup entry point or an assumed order between scripts. This makes it difficult to determine the exact type of each item in a table at a given moment, as the type of the identifiers on which the item depends may not be determined. It is possible that the type may be determined at some point in the future, or it may never be determined. In this case, how should we assume the type of this table?

georgiy-belyanin: I see the problem you mean now. Thank you

I suggest a compromise: the code may behave as follows.

* If LSP sees only one non-unknown type `T` within the array table it infers the table as `T[]` array.
* If LSP sees multiple non-unknown types and a few unknown elements within the array table it infers it as a tuple of known types and unknown types.
* If LSP sees `<=10` no unknown types within the array table it infers the table as a tuple of known types.
* If LSP sees `>10` types it infers `TableConst` type.

```lua
-- Unknown
local a
---@type number
local b
---@type number
local c
---@type string
local d

--- number[] is inferred, a is assumed to be a number
local f = {a, b, c}

--- (number, string, unknown)
local g = {c, d, a}

--- (number, string)
local h = {c, d}
```

CppCXY: In my current inference system, I perform loop expression inference through the unresolve phase until a fixed point is reached. If it still cannot be inferred at the fixed point, it should enter the force type inference phase (I have not implemented this logic), and what you mentioned should be the content of this phase.

CppCXY:
Please resolve conflicts

CppCXY:
I added support myself.

PR #50, opened 2025-02-08
add function overload completions

georgiy-belyanin:
This patch adds completions for functions overloads. Basically it just adds extra function signatures.

For instance, when you write `table.ins...` you receive the following completions.
```
insert FUNCTION (list, pos, value)number
insert FUNCTION (list, value)number
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 `NEW` Support type-check when casting tuples to arrays.
 
+`NEW` Now autocompletion suggests function overloads.
+
 # 0.4.6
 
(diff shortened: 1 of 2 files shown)


xuhuanzy:
This may require fallback, because `luals` rules allow overloading function fields, which can cause the completion list to be too long.
We should use `alt + up arrow/down arrow` to switch method prompts

xuhuanzy:
I will do some work to make it automatically infer the matching function based on parameter input, which is also consistent with the behavior of `luals`. After completion, should this function be removed? Or provide a configuration item to enable it?

PR #183, opened 2025-03-11
infer for range vars from aliases and overloads

georgiy-belyanin:
This patch adds support for deriving for-range loop variables from the
iterator types specified as aliases and class overloads.

I.e. you may alias `fun(v: any): K, V` and if you try to iterate through
the instance of this type the loop variables will be inferred as `K, V`.
If the class provide a function overload it's also used for type
inference.

It might be helpful for writing complex functional libraries for
determining custom iterators.

Example.

```lua
---@alias foo (fun(tbl: any): (number, integer))

---@class bar
---@overload fun(tbl: any): (number, integer)

---@type foo
local b = {}

---@type bar
local c = {}

for k, v in b do
    -- k is inferred as number
    -- v is inferred as string
end

for k, v in c do
    -- k is also inferred as number
    -- v is also inferred as string
end
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 `NEW` Compact luals fun return syntax like: `(name: string, age: number)`
 
+`NEW` Aliases and overloads of iterator functions (i.e `fun(v: any): (K, V)` where `K` is the key type and `V` is the value type) are now used to infer types in `for` loops
+
 # 0.5.4
 
 `Fix` Fix generic dots params type check
--- a/crates/emmylua_code_analysis/src/compilation/analyzer/lua/stats.rs
+++ b/crates/emmylua_code_analysis/src/compilation/analyzer/lua/stats.rs
@@ -7,7 +7,7 @@ use crate::{
     compilation::analyzer::unresolve::{
         merge_decl_expr_type, merge_member_type, UnResolveDecl, UnResolveIterVar, UnResolveMember,
     },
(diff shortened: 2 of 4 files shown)


CppCXY:
why not use `---@alias foo fun(tbl: any): (number, string)`

georgiy-belyanin:
Actually, I've implemented it because I work on a project using a LuaJIT specific functional library and the class both contains fields and might be used as an iterator.

But you're right it may also be handy. I've implemented inferring type from the aliases. Force-pushed. Check this out, please.

CppCXY:
According to LuaLS, if a class is callable, it should use ---@overload.
```lua
---@class AAA
---@overload fun(param1: string): AAA


local a---@type AAA
a("1313")
```

CppCXY:
if you use inherit, you will meet the bad usage:
```lua
---@alias callable fun(param1: string): AAA
---@class AAAA: callable
```

CppCXY:
And I cannot expand it during analysis, because I meet such bad usage (Direct or indirect recursive definition):
```lua
---@alias callable callable
---@class AAAA: callable
```

georgiy-belyanin:
Thank you for detailed comments. Let's then just infer the variables from the aliases.

CppCXY:
> Thank you for detailed comments. Let's then just infer the variables from the aliases.

I think you can support overload, you can get overload result from operator

georgiy-belyanin:
Applied.

CppCXY:
can you add test for this case?

georgiy-belyanin:
Added a test case. Force-pushed.

PR #211, opened 2025-03-17
Diagnostics for unnecessary assert

ribru17:
Warns on unnecessary assertions (on expressions that always evaluate to `true`):

image

image

I might need help with tests and i18n. Also, this does not work for aliases, I couldn't figure out why:

image

### DIFF
--- a/crates/emmylua_code_analysis/src/db_index/type/types.rs
+++ b/crates/emmylua_code_analysis/src/db_index/type/types.rs
@@ -272,6 +272,19 @@ impl LuaType {
         }
     }
 
+    pub fn is_always_truthy(&self) -> bool {
+        match self {
+            LuaType::Nullable(_)
+            | LuaType::Nil
+            | LuaType::Boolean
+            | LuaType::Any
+            | LuaType::Unknown => false,
+            LuaType::BooleanConst(boolean) | LuaType::DocBooleanConst(boolean) => boolean.clone(),
+            LuaType::Union(u) => u.types.iter().all(|t| t.is_always_truthy()),
+            _ => true,
+        }
+    }
+
     pub fn is_tuple(&self) -> bool {
         matches!(self, LuaType::Tuple(_))
     }
--- a/crates/emmylua_code_analysis/src/diagnostic/checker/unnecessary_assert.rs
+++ b/crates/emmylua_code_analysis/src/diagnostic/checker/unnecessary_assert.rs
@@ -1,0 +1,39 @@
+use emmylua_parser::{LuaAstNode, LuaCallExpr};
+
+use crate::{DiagnosticCode, SemanticModel};
+
+use super::DiagnosticContext;
+
+pub const CODES: &[DiagnosticCode] = &[DiagnosticCode::UnnecessaryAssert];
+
+pub fn check(context: &mut DiagnosticContext, semantic_model: &SemanticModel) -> Option<()> {
+    let root = semantic_model.get_root().clone();
+    for call_expr in root.descendants::<LuaCallExpr>() {
+        if call_expr.is_assert() {
+            check_assert_rule(context, semantic_model, call_expr);
+        }
+    }
+
+    Some(())
+}
+
+fn check_assert_rule(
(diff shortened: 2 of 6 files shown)


crates/emmylua_code_analysis/src/db_index/type/types.rs
```diff
@@ -272,6 +272,19 @@ impl LuaType {
         }
     }
 
+    pub fn is_always_truthy(&self) -> bool {
```
CppCXY, line 275: I think if this function is implemented solely for this check, it would be better to write it within the check, so that the relevance is stronger.

crates/emmylua_code_analysis/src/diagnostic/checker/unnecessary_assert.rs
```diff
@@ -0,0 +1,39 @@
...
+) -> Option<()> {
+    let args = call_expr.get_args_list()?;
+    let arg_exprs = args.get_args().collect::<Vec<_>>();
+    if let Some(first_expr) = arg_exprs.first() {
+        let expr_type = semantic_model.infer_expr(first_expr.clone());
+        if expr_type.is_some_and(|t| t.is_always_truthy()) {
+            context.add_diagnostic(
+                DiagnosticCode::UnnecessaryAssert,
+                call_expr.get_range(),
+                t!("Unnecessary assert: this expression is always truthy").to_string(),
```
CppCXY, line 33: i18n can write in https://github.com/CppCXY/emmylua-analyzer-rust/blob/main/crates/emmylua_code_analysis/locales/lint.yml

CppCXY:
> I might need help with tests and i18n. Also, this does not work for aliases, I couldn't figure out why:

 Because is_assert is only computed during the parsing phase through text matching, supporting aliases generally doesn't have much significance.

crates/emmylua_code_analysis/src/db_index/type/types.rs
ribru17: After trying this, it looks like it cannot be done since `types` is a private field; is it alright to leave it here? I have noticed some other functions are only used once, or not at all (`is_tuple`), and I believe this helper will be useful in other places in the future, like for unnecessary `if` statements

CppCXY:
please resolve the conflict

crates/emmylua_code_analysis/src/db_index/type/types.rs
CppCXY: ok, I accept

crates/emmylua_code_analysis/src/diagnostic/checker/unnecessary_assert.rs
ribru17: I just used google translate for this- let me know if there are inaccuracies :laughing:

CppCXY: This expression isn't very idiomatic in Chinese; I suggest you use GPT for translation, just as our conversation is also translated through GPT

ribru17: Better now? If not you may need to do it yourself, apologies

CppCXY: zh_HK refers to translations between Simplified and Traditional Chinese, not a regional language. However, I know you might not understand the difference, so let's leave it as is; I'll make changes later.

ribru17: Oh, sorry. I thought it was cantonese

CppCXY:
can you add a test for this check?

ribru17:
Sure, I can do this in a couple days

lewis6991:
I think this still needs a bit of work:

image

lewis6991:
Ok. This check is falsely triggering in quite a lot of places in gitsigns.nvim. I'll raise a separate issue unless one already exists.

PR #255, opened 2025-03-23
Refactor return type mismatch check

ribru17:
It usually doesn't make sense to return `Option<()>`


CppCXY:
please migrate to https://github.com/CppCXY/emmylua-analyzer-rust/pull/256

PR #256, opened 2025-03-23
Refactor: migrate diagnostics to use trait rather than macro

ribru17:
This commit makes it so that diagnostic checkers no longer have to define magic public variables which will be called by a macro. Instead, they all implement a common trait which will easily enforce that all the checking data is defined. This also means compile errors will be quicker and more understandable.

### DIFF
--- a/crates/emmylua_code_analysis/src/diagnostic/checker/analyze_error.rs
+++ b/crates/emmylua_code_analysis/src/diagnostic/checker/analyze_error.rs
@@ -1,24 +1,25 @@
 use crate::{DiagnosticCode, SemanticModel};
 
-use super::DiagnosticContext;
+use super::{Checker, DiagnosticContext};
 
-pub const CODES: &[DiagnosticCode] = &[
-    DiagnosticCode::TypeNotFound,
-    DiagnosticCode::AnnotationUsageError,
-];
+pub struct AnalyzeErrorChecker;
 
-pub fn check(context: &mut DiagnosticContext, _: &SemanticModel) -> Option<()> {
-    let db = context.get_db();
-    let file_id = context.get_file_id();
-    let diagnostic_index = db.get_diagnostic_index();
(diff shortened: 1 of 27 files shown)


CppCXY:
When I first designed the checker, I essentially structured it in this way. The current question is whether I should redesign the checker in an oo style, and if there are other linter examples that make this change more compelling.

ribru17:
In my opinion it makes more sense this way, but it is your decision of course. The previous macro way was essentially just manually specifying what the compiler would do anyway (with a trait)

crates/emmylua_code_analysis/src/diagnostic/checker/analyze_error.rs
```diff
@@ -1,24 +1,25 @@
...
+        DiagnosticCode::AnnotationUsageError,
+    ];
 
-    Some(())
+    fn check(context: &mut DiagnosticContext, _: &SemanticModel) {
+        let db = context.get_db();
+        let file_id = context.get_file_id();
+        let diagnostic_index = db.get_diagnostic_index();
+        let Some(diagnostics) = diagnostic_index.get_diagnostics(&file_id) else {
+            return;
```
CppCXY, line 18: Why remove Option<()>? That way I can’t '?'

CppCXY:
I referred to Clippy, which has a very similar structure, and perhaps this makes sense.
https://github.com/rust-lang/rust-clippy/blob/master/clippy_lints/src/lifetimes.rs

crates/emmylua_code_analysis/src/diagnostic/checker/analyze_error.rs
ribru17: True, but generally returning nothing is better because option introduces added complexity (must use `unwrap`, `is_some`, etc.). It doesn't make sense just to add a pointless option return value for this syntax, especially since we can just return early like in the above. Option should be used if it makes semantic sense

PR #284, opened 2025-03-27
fix: stdlib remove some overloads

lewis6991:
Not sure why these were added, but they shouldn't be necessary.

### DIFF
--- a/crates/emmylua_code_analysis/resources/std/global.lua
+++ b/crates/emmylua_code_analysis/resources/std/global.lua
@@ -73,8 +73,7 @@ function collectgarbage(opt, ...) end
 --- (`stdin`). Returns all values returned by the chunk. In case of errors,
 --- `dofile` propagates the error to its caller (that is, `dofile` does not run
 --- in protected mode).
----@overload fun():table
----@param filename string
+---@param filename? string
 ---@return table
 function dofile(filename) end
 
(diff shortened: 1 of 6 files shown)


xuhuanzy:
This is a case of "the ancestral laws must not be altered" (a historical legacy issue). The old version of `emmylua` had effectively ceased maintenance long ago (the original author stopped updating it), and these `std` definitions have long been outdated.  

Although it's still called `emmylua` now, the core is entirely different. In fact, I'm quite puzzled as to why the project continues to bear the name `emmylua`.

lewis6991:
I agree about the name, especially since this project supports [LuaCATS](https://luals.github.io/wiki/annotations/) and not [EmmyLua](https://emmylua.github.io/annotation.html)

CppCXY:
Because I was an member of EmmyLua and the early version of this project was implemented in [C#](https://github.com/CppCXY/EmmyLuaAnalyzer), the name was inherited

lewis6991:
Would it make sense to change the name?

CppCXY:
which name? I want to keep the name 'emmylua' because 'luals' is already taken, and everyone is very accustomed to this name. Additionally, it is called 'analyzer' because it was originally copied from 'rust-analyzer'.

lewis6991:
`emmylua` doesn't make sense since this project doesn't support [EmmyLua](https://emmylua.github.io/annotation.html). 

Maybe `lua_analyzer` or `luacats_ls`?

xuhuanzy:
In fact, the functionality of that plugin is quite primitive and includes some unnecessary features.  
Additionally, there are now noticeable differences between the plugin and LuaCATS.

xuhuanzy:
The documentation will be provided later, but the foundational work is not yet complete, so providing it now is unnecessary.  
Of course, it would be best if new maintainers could join in to help with the documentation.
I recommend using `VitePress` to build the documentation.

lewis6991:
Basic markdown files within this project would be sufficient for the short term.

PR #337, opened 2025-04-07
fix(std): define math.huge as integer

lewis6991:
`math.floor(math.huge) == math.huge` so should define it as an integer.

### DIFF
--- a/crates/emmylua_code_analysis/resources/std/math.lua
+++ b/crates/emmylua_code_analysis/resources/std/math.lua
@@ -87,7 +87,7 @@ function math.fmod(x, y) end
 
 ---
 --- The float value `HUGE_VAL`, a value larger than any other numeric value.
----@type number
+---@type integer
 math.huge = nil
 
 ---


taminomara:
Sorry but this is wrong, `math.huge` is a floating point infinity value, it should not be an integer:

```lua
math.tointeger(math.huge) == nil
```

lewis6991:
Then please explain `math.floor(math.huge) == math.huge`

If this isn't true then `math.floor` doesn't return integers.

taminomara:
> Then please explain math.floor(math.huge) == math.huge
>
> If this isn't true then math.floor doesn't return integers.

It's true, if `math.floor` gets a floating point number, then it returns a rounded down floating point number. That includes all caveats regarding special values for floating points (inf, nan, etc.) That also holds true for cases when the given value does not fit a 64-bit signed integer:

```lua
math.floor(2 ^ 64) == 1.844674407371e+19
```

So, yes, technically `math.floor` should return a `number`.

taminomara:
I'm not sure how to annotate `math.floor`, though. If we want to force users to check for these corner cases, then we should go with `number`. If we prefer it to be more ergonomic, then we should go with `integer` (most users will not run into these corner cases, probably).

Issue #248, opened 2025-03-23
should `string[]` elements be nilable?


lewis6991:
```lua
local a --- @type string[]
local b, c = unpack(a)
-- type string, string
-- should it actually be string?, string?
```

For context this is because of this code:

<img width="827" alt="Image" src="https://github.com/user-attachments/assets/cf9fc839-4672-49f7-b035-c73c1abd4544" />

If we decide the current behaviour should stay the same, then I'll need to add a `@diagnostic disable` on this assert. Alternatively `vim.split` could return `(string?)[]`

lewis6991:
Another situation:

```lua
local stdout --- @type string[]

local line1 = assert(stdout[1]) -- causes unnecessary assert diagnostic
```

We don't know the length of `stdout`, so any index may be `nil`.

lewis6991:
array indexes are now nilable

PR #371, opened 2025-04-17
feat: treat array index as nil-able

ribru17:
Expressions like this:

```lua
---@type integer[]
local arr = {1, 2, 3}
return arr[2]
```

should be considered as type `integer|nil`, since array accesses can always *potentially* return `nil`. This commit marks those expressions as such.

Related to #248

### DIFF
--- a/crates/emmylua_code_analysis/src/semantic/infer/infer_index.rs
+++ b/crates/emmylua_code_analysis/src/semantic/infer/infer_index.rs
@@ -167,12 +167,17 @@ fn infer_array_member(
     index_expr: LuaIndexMemberExpr,
 ) -> Result<LuaType, InferFailReason> {
     let key = index_expr.get_index_key().ok_or(InferFailReason::None)?;
+    let expression_type = if db.get_emmyrc().strict.array_index {
+        TypeOps::Union.apply(array_type, &LuaType::Nil)
+    } else {
+        array_type.clone()
+    };
     match key {
-        LuaIndexKey::Integer(_) => Ok(TypeOps::Union.apply(array_type, &LuaType::Nil)),
+        LuaIndexKey::Integer(_) => Ok(expression_type),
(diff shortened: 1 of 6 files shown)


crates/emmylua_code_analysis/src/semantic/infer/infer_index.rs
```diff
@@ -668,12 +666,12 @@ fn infer_member_by_index_array(
...
     let member_key = index_expr.get_index_key().ok_or(InferFailReason::None)?;
     if member_key.is_integer() {
-        return Ok(base.clone());
+        return Ok(TypeOps::Union.apply(base, &LuaType::Nil));
     } else if member_key.is_expr() {
         let expr = member_key.get_expr().ok_or(InferFailReason::None)?;
         let expr_type = infer_expr(db, cache, expr.clone())?;
         if check_type_compact(db, &LuaType::Number, &expr_type).is_ok() {
-            return Ok(base.clone());
+            return Ok(TypeOps::Union.apply(base, &LuaType::Nil));
```
ribru17, line 674: I am assuming this is also necessary?

ribru17:
cc @lewis6991, will this be too annoying in practice?

CppCXY:
I think we need a configuration for strictly indexing arrays, because I noticed that even in Rust, array access operations do not return Option, and other languages, including TypeScript, do not return T | null.

lewis6991:
#248

ribru17:
I've gated this change behind a config option, defaulting to `true`. My reasoning for having this be opt-out is that there exists `pairs` and `ipairs` for safe array iteration, and indexing arrays manually (**important:** talking about *arrays* specifically, not a tuple like `[integer, integer]`) is not common enough that it warrants allowing this extra unsafety, in my opinion. Fine with having this as opt-in if there is another argument, of course

PR #385, opened 2025-04-19
feat: diagnostics for unnecessary "if" conditions

ribru17:


crates/emmylua_code_analysis/src/diagnostic/checker/unnecessary_if.rs
```diff
@@ -0,0 +1,54 @@
...
+    }
+}
+
+fn check_condition(
+    context: &mut DiagnosticContext,
+    semantic_model: &SemanticModel,
+    condition: LuaExpr,
+) -> Option<()> {
+    let expr_type = semantic_model.infer_expr(condition.clone()).ok()?;
+    let first_type = match &expr_type {
```
CppCXY, line 33: Generally, you don't need to worry about multiple return values. Functions with multiple return values will determine whether to return multiple values based on whether the item is the last item in the list when returning.

ribru17: Hm, but wouldn't we still have to check for this scenario?

```lua
---@return false, integer?
local function test()
  return false, nil
end

if test() then
  -- ...
end
```
I believe the condition always checks the first return value, I could be wrong though

CppCXY: https://github.com/EmmyLuaLs/emmylua-analyzer-rust/blob/37b9eb87b10eac6ae2cf34a2ff05a36532181f61/crates/emmylua_code_analysis/src/semantic/infer/infer_call/mod.rs#L416

PR #386, opened 2025-04-19
ci: ensure that the schema has been generated

ribru17:


.github/workflows/build.yml
```diff
@@ -135,3 +135,23 @@ jobs:
             emmylua_ls-darwin-x64.tar.gz
             emmylua_ls-darwin-arm64.tar.gz
           token: ${{ secrets.RELEASE }}
+  generate-schema:
```
CppCXY, line 138: why not add to test.yml

PR #422, opened 2025-05-05
fix: string.match and string.find

lewis6991:


CppCXY:
need I support this  grammar

lewis6991:
Sorry, I don't understand?

CppCXY:
Previously, I only supported the format "<name> ...."

Issue #435, opened 2025-05-11
`emmylua_doc_cli`: add option to export JSON data only for custom templating


rhys-vdw:
We would like to use our own templating system to generate docs so that we can use custom link Lua types to their doc pages.

Would it be possible to export a single JSON file like so:

```console
emmylua_doc_cli --json
```

```jsonc
{
  "globals": [
    {
      "namespace": "Foo",
      "name": "Bar",
      "display": "Foo.Bar = {\n  x = integer\n}",
      "property": { "description": "" },
      "fields": [{ /* ... */ }]
    }
  ],
  "types": []
}
```

CppCXY:
The Lua files are already automatically generated by you. If you still want to generate content based on the doc_cli, why not directly generate your documentation from C++?

rhys-vdw:
Because we have plans to document our Lua codebase too, and currently we have Lua definition files too.

I can make a PR for this, I don't think it will be hard.

CppCXY:
you can make a pr for this

taminomara:
Hi! I'm an author of [Sphinx-Lua-La], and I'm also interested in this implemented. Is anyone working on a PR for it? If no, I might look into it.

[Sphinx-Lua-La]: https://taminomara.github.io/sphinx-lua-ls/

xuhuanzy:
Not currently. There are two maintainers for this project, but our primary focus is not on `emmylua_doc_cli`. We would greatly appreciate contributions for additional output formats for emmylua_doc_cli, like JSON, VitePress, and others, if you are able to provide them.

rhys-vdw:
> Hi! I'm an author of [Sphinx-Lua-La](https://taminomara.github.io/sphinx-lua-ls/), and I'm also interested in this implemented. Is anyone working on a PR for it? If no, I might look into it.

I had a look at the code yesterday, it is not _completely_ trivial and would like to get to it eventually, but I'd certainly appreciate if you tackled it. We definitely need it. ❤

taminomara:
You can assign this to me then, I'll take a look during weekend 😺

PR #456, opened 2025-05-15
JSON format for documentation export

taminomara:
Partial solution for #435.

The only thing that's missing from export is descriptions for modules. They are not stored in the database as of now. I could add them, but I'm not sure where to store them. I guess `VFS` or `LuaModuleIndex`? @CppCXY any ideas?

### DIFF
--- a/crates/emmylua_doc_cli/src/json_generator/export.rs
+++ b/crates/emmylua_doc_cli/src/json_generator/export.rs
@@ -1,0 +1,330 @@
+use crate::common::{render_const, render_typ};
+use crate::json_generator::json_types::*;
+use emmylua_code_analysis::{
+    DbIndex, FileId, LuaDeprecated, LuaMemberKey, LuaMemberOwner, LuaNoDiscard,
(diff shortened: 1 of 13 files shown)


CppCXY:
from `ModuleInfo.property_owner_id` and query from propertyIndex

taminomara:
So, extend `LuaSemanticDeclId` to support module IDs?

crates/emmylua_doc_cli/src/json_generator/export.rs
```diff
@@ -0,0 +1,337 @@
...
+    let global_index = db.get_global_index();
+    let module_index = db.get_module_index();
+    let type_index = db.get_type_index();
+    let vfs = db.get_vfs();
+    let globals = global_index.get_all_global_decl_ids();
+
+    globals
+        .into_iter()
+        .filter(|global| {
+            !(module_index.is_std(&global.file_id) || module_index.is_library(&global.file_id))
```
CppCXY, line 96: should use a common method?

PR #457, opened 2025-05-15
Clean up some std definitions

### LINKED ISSUE #380: Some standard library functions are mistyped
crates/emmylua_code_analysis/resources/std/global.lua#L247:
```lua
---@generic T, R, R1
---@param f fun(...: T...): R1, R...
---@param ... T...
---@return boolean, R1|string, R...
function pcall(f, ...) end
```
assumes that errors must be strings, but for example, `type( pcall(error, {}) )` evaluates to `"table"`.`

```lua
-- crates/emmylua_code_analysis/resources/std/global.lua#L117:
---@generic V
---@param t V[] | table<any, V> | {[any]: V}
---@return fun(tbl: any):int, V
function ipairs(t) end

-- crates/emmylua_code_analysis/resources/std/global.lua#L235:
---@generic K, V
---@param t table<K, V> | V[] | {[K]: V}
---@return fun(tbl: any):K, V
function pairs(t) end
```
Are each missing return values, as 
- `ipairs(t)` returns an iterator function, `t`, and `0`
- `pairs(t)` returns `next`, `t`, and `nil` (though the `nil` can only be observed with `select('#', pairs{})`)
Also, the iterator functions should have their first return value be nilable (`K?`/`int?`), as a `nil` return is used to indicate the end of an iterator (for loops already handle this by removing the nilability).

taminomara:
Fix #380 (and possibly some others)

### DIFF
--- a/crates/emmylua_code_analysis/resources/std/builtin.lua
+++ b/crates/emmylua_code_analysis/resources/std/builtin.lua
@@ -17,13 +17,13 @@
 -- Built-in Types
 
 ---
---- The type *nil* has one single value, **nil**, whose main property is to be 
---- different from any other value; it usually represents the absence of a 
+--- The type *nil* has one single value, **nil**, whose main property is to be
+--- different from any other value; it usually represents the absence of a
 --- useful value.
 ---@class nil
 
 ---
---- The type *boolean* has two values, **false** and **true**. Both **nil** and 
+--- The type *boolean* has two values, **false** and **true**. Both **nil** and
 --- **false** make a condition false; any other value makes it true.
 ---@class boolean
 
@@ -44,22 +44,22 @@
 ---@class integer
 
 ---
---- The type *userdata* is provided to allow arbitrary C data to be stored in 
---- Lua variables. A userdata value represents a block of raw memory. There 
+--- The type *userdata* is provided to allow arbitrary C data to be stored in
+--- Lua variables. A userdata value represents a block of raw memory. There
 --- are two kinds of userdata: *full userdata*, which is an object with a block
 --- of memory managed by Lua, and *light userdata*, which is simply a C pointer
---- value. Userdata has no predefined operations in Lua, except assignment 
---- and identity test. By using *metatables*, the programmer can define 
+--- value. Userdata has no predefined operations in Lua, except assignment
+--- and identity test. By using *metatables*, the programmer can define
 --- operations for full userdata values. Userdata values cannot be
---- created or modified in Lua, only through the C API. This guarantees the 
+--- created or modified in Lua, only through the C API. This guarantees the
 --- integrity of data owned by the host program.
 ---@class userdata
 
 ---@class lightuserdata
 
 ---
---- The type *thread* represents independent threads of execution and it is 
---- used to implement coroutines. Lua threads are not related to 
+--- The type *thread* represents independent threads of execution and it is
+--- used to implement coroutines. Lua threads are not related to
 --- operating-system threads. Lua supports coroutines on all systems, even those
 --- that do not support threads natively.
 ---@class thread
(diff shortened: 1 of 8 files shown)


crates/emmylua_code_analysis/resources/std/builtin.lua
```diff
@@ -125,3 +123,9 @@
 ---@alias collectgarbage_opt std.collectgarbage_opt
 
 ---@alias metatable std.metatable
+
+---@alias int integer
+
+---@alias str string
+
+---@alias bool boolean
```
taminomara, line 131: Note: I'm not sure about adding these. Personally, I'd rather had compact types under a feature flag, disabled by default.

crates/emmylua_code_analysis/resources/std/global.lua
```diff
@@ -243,7 +250,7 @@ function pairs(t) end
 ---@generic T, R, R1
 ---@param f fun(...: T...): R1, R...
 ---@param ... T...
----@return boolean, R1|string, R...
+---@return boolean, R1|any, R...
```
taminomara, line 253: Here and in `xpcall`, it would be better to have a separate type like `unknown`, because `any` is too broad.

CppCXY:
almost modify is wrong, Please do not modify my basic definitions of various generic functions, as these logics have already been handled during matching. Also, do not change the places where I use "integer" to "number," and do not add the "str" type.

taminomara:
Why? 

Why do you want to have `int` but not `str`? This doesn't make sense.

Changing integer to number is required, because those annotations are incorrect. `math.huge` is not an integer, it is a floating point infinity.

CppCXY:
> Why?
> 
> Why do you want to have `int` but not `str`? This doesn't make sense.

Because common programming languages use "string" as the type for strings rather than the shorthand "str," but "int" and "bool" are very commonly used across various programming languages.

> Changing integer to number is required, because those annotations are incorrect. `math.huge` is not an integer, it is a floating point infinity.

I need time to study further before deciding whether it should be "number." Additionally, I am primarily focusing on the functions math.floor and math.ceil. Why should it be "number"?

taminomara:
> Because common programming languages use "string" as the type for strings rather than the shorthand "str," but "int" and "bool" are very commonly used across various programming languages.

Common programming languages are consistent. They either define all types without abbreviations, or all types with abbreviations. They don't mix abbreviated and non-abbreviated types, like `string` and `int`.

Python and Rust, for example, prefer abbreviations: `str`, `bool`, `int` (`i32` & co. in Rust).

Typescript and Java, on the other hand, prefer full names: `String`, `Number`.

A lot of other languages use abbreviations for built-in (i.e. simple) types, and non-abbreviated names for library types. Examples include C++, C#, Haskell, Rust's `String` vs `&str`, and many more. If you wish to follow this convention, then Lua's string should be names `str`, because it is a built-in type.

> Additionally, I am primarily focusing on the functions math.floor and math.ceil. Why should it be "number"?

See my comments in #337. `math.floor` and `math.ceil` return integers in cases when their argument doesn't fit signed 64 bit integer, or not a number at all:

```lua
math.floor(2 ^ 64) --> 1.844674407371e+19
math.type(math.floor(2 ^ 64)) --> float

math.floor(math.huge) --> inf
math.type(math.floor(math.huge)) --> float

math.floor(0 / 0) --> nan
math.type(math.floor(0 / 0)) --> float
```

Although we can keep them as integers for convenience.

taminomara:
> I need time to study further before deciding whether it should be "number."

Lua itself tells you it's a float. Why would you define it as an integer?

```lua
math.type(math.huge) -> float
math.tointeger(math.huge) -> nil
```

CppCXY:
from the lua c code:
image

Essentially, it calls the C language floor function. Although the floor function itself returns a double, when Lua returns the result, it decides whether to return an integer or a float based on whether the value is between INT_MAX and INT_MIN. Therefore, we cannot arbitrarily assume that its return value is always an integer or a number. However, I think this behavior is mainly to ensure that there is always a value to return, and the return value tends to be an integer. So I still prefer to treat its return value as an integer.

CppCXY:
image
math.huge is the overflow value INF for double, and according to IEEE-754, it is still a double.
https://en.wikipedia.org/wiki/IEEE_754#Overflow%EF%BC%89
@lewis6991

crates/emmylua_code_analysis/resources/std/global.lua
```diff
@@ -435,20 +442,21 @@ _VERSION = "Lua 5.4"
...
----@generic T
----@param i? integer
----@param j? integer
----@param list [T...]
----@return T?...
+---@generic T, I: integer, J: integer
+---@param list [T...] | T[] | table<integer, T> | {[integer]: T}
+---@param i? I
+---@param j? J
+---@return T...
```
CppCXY, line 457: donot change this: because it need implement in special type std.unpack

crates/emmylua_code_analysis/resources/std/table.lua
```diff
@@ -97,20 +98,21 @@ function table.remove(list, pos) end
...
----@generic T
----@param list [T...] | T[] | table<any, T>
----@param i? integer
----@param j? integer
+---@generic T, I: integer, J: integer
+---@param list [T...] | T[] | table<integer, T> | {[integer]: T}
+---@param i? I
+---@param j? J
 ---@return T...
+--XXX: @return std.Unpack<T..., I, J>
```
CppCXY, line 115: same donot change

CppCXY:
Most of the changes have already been implemented by other PRs; the small remaining parts I will choose to implement myself.

PR #513, opened 2025-05-30
emmylua_check: print log warnings and errors to stderr

lewis6991:
- Add `--verbose` flags.
- Improve docs for `--ignore` and `--config`.
- Add logging for include/excludes.


lewis6991:
Summaries are already printed per file so I think it makes sense to add at the end.

I'm not really sure if _just_ having a summary is a good idea, can't you just pipe to `grep '^[WEIH]'` or `tail -n 5`?

crates/emmylua_check/src/init.rs
```diff
@@ -4,9 +4,30 @@ use emmylua_code_analysis::{
...
 };
 
+fn root_from_configs(config_paths: &Vec<PathBuf>, fallback: &PathBuf) -> PathBuf {
+    if config_paths.len() > 1 {
+        fallback.clone()
+    } else {
+        let config_path = &config_paths[0];
+        // Need to convert to canonical path to ensure parent() is not an empty
+        // string in the case the path is a relative basename.
+        match config_path.canonicalize() {
```
lewis6991, line 14: @clason this should hopefuly fix the issue you were seeing.

clason:
> Summaries are already printed per file so I think it makes sense to add at the end.
> 
> I'm not really sure if _just_ having a summary is a good idea, can't you just pipe to `grep '^[WEIH]'` or `tail -n 5`?

Yeah, it's just convenience. (`tree-sitter parser` works like that, and I quite like that I can tune the output to what I want to see precisely.) 

Basically, emmyluals output is quite verbose (with 7-8 lines per diagnostic), so sometimes I just want a more compact option if I simply want to quickly check for regressions.

But no big deal either way.

clason:
Now `emmylua_check -c .luarc.json .` fails with 
```
error: The following required arguments were not provided:
    <workspace>
```

lewis6991:
> It also always prints 
> ```
> ERROR: Failed to read config file: "/Users/clason/build/nvim-treesitter/./.emmyrc.json", error: File not found or unreadable
> ```
> even if it has found a `.luarc.json`.

It was always doing that when `-c` wasn't provided, we just print the error now. Maybe we should check the existence first.

clason:
> Now `emmylua_check -c .luarc.json .` fails with
> 
> ```
> error: The following required arguments were not provided:
>     <workspace>
> ```
> 
> (Seems to be happening as soon as I provide a `-c`.)

Oh, now the argument order is different (and mandatory), it needs to be `<workspace> -c <file>`; `-c <file> <workspace>` no longer works.

lewis6991:
> Oh, now the argument order is different (and mandatory), it needs to be <workspace> -c <file>; -c <file> <workspace> no longer works.

Yeah just noticed that too. Not sure why that's required. This is all handled by a library, so need to investigate why it changed.

lewis6991:
Using `-c=<file>` also works.

lewis6991:
> Now `emmylua_check -c .luarc.json .` fails with
> 
> ```
> error: The following required arguments were not provided:
>     <workspace>
> ```
> 
> (Seems to be happening as soon as I provide a `-c`.)

Now that config is a vector, I think we are stuck with this new behaviour. The workaround is to use `-c=file` instead of `-c file`.

The library structopt that we use for cli args is basically deprecated so we need to upgrade to clap to fix the behaviour.

PR #523, opened 2025-06-04
refactor flow algorithm

CppCXY:


clason:
Sorry to comment here, but after this got merged I see a performance regression when running `emmylua_check` on the top-level directory of a repo: The actual lua files get checked quickly (no regression there), but then it just "freezes" for a second or two before exiting (with code 1). Checking only the `lua` subdirectory is fast (although there might also be a delay, it's too short to measure).

Do you have any suggestion how to debug this?

CppCXY:
> Sorry to comment here, but after this got merged I see a performance regression when running `emmylua_check` on the top-level directory of a repo: The actual lua files get checked quickly (no regression there), but then it just "freezes" for a second or two before exiting (with code 1). Checking only the `lua` subdirectory is fast (although there might also be a delay, it's too short to measure).
> 
> Do you have any suggestion how to debug this?

which repo?

clason:
https://github.com/nvim-treesitter/nvim-treesitter/tree/main (`main` branch)

with a local `.emmyluarc.json` on top:
```json
{
  "workspace" : {
    "ignoreDir": [ "lua/nvim-treesitter/async.lua" ]
  },
  "diagnostics" : {
    "disable" : [
      "unnecessary-if",
      "unnecessary-assert",
      "duplicate-require"
    ]
  },
  "strict": {
    "typeCall": true,
    "arrayIndex": true
  }
}
```

CppCXY:
I’ll test it tomorrow when I have time. I’m getting off work now.

clason:
Now the check is instant again! Thanks for the very fast fix!

Issue #580, opened 2025-07-07
libraries for `emmylua_check`, `emmylua_ls` and `emmylua_doc`


mrcjkb:
Hey 👋 

Some friends and I are working on [lux](https://github.com/nvim-neorocks/lux), a package manager for Lua.

It would be amazing if we could add `lx check` and `lx lsp` command wrappers to `lux-cli`, that would delegate to `emmylua_check` and `emmylua_ls` (and perhaps eventually also one for `emmylua_doc_cli`):

- https://github.com/nvim-neorocks/lux/issues/558

If you provided a library API for those crates, this should be quite trivial to implement.
We've already done something similar for [stylua](https://github.com/JohnnyMorganz/StyLua) with a `lx fmt` command.

CppCXY:
I have add the lib.rs to emmylua_check, you can try first. I'm not sure if it's possible to directly reference GitHub as a crate, or maybe I'll update it to crates.io on Friday.

CppCXY:
I have add lib.rs to `emmylua_check`, `emmylua_doc_cli`, `emmylua_ls` crates. But if you import all crates, your final binary file might be very large.

mrcjkb:
Thanks for the quick response 🙏

> I have add lib.rs to `emmylua_check`, `emmylua_doc_cli`, `emmylua_ls` crates. But if you import all crates, your final binary file might be very large.

Hmm, I hadn't thought about that. How large are the CLI binaries? 
Could adding a feature flag for the CLI help reduce the size?

I suppose `emmylua_ls` is the largest one? Perhaps adding that to lux is overkill anyway.

PR #589, opened 2025-07-08
feat(emmylua_check): library

mrcjkb:
This PR introduces a `cli` feature flag (enabled by default) to `emmylua_check`.
If disabled, it removes the `clap` dependency. This should allow the use of `emmylua_check` as a library in CLI applications that use a conflicting version of clap (see #580).

If you agree with this approach, I'll look into implementing the same for the other CLI tools.


CppCXY:
Please fix test

mrcjkb:
:thinking: hmm, this test failure seems unrelated. ~Is it passing on `main`?~
Edit: It also fails [here](https://github.com/EmmyLuaLs/emmylua-analyzer-rust/actions/runs/16165084686/job/45624989996) on `main`.

```
failures:

---- handlers::test::hover_function_test::tests::test_3 stdout ----

thread 'handlers::test::hover_function_test::tests::test_3' panicked at crates/emmylua_ls/src/handlers/test/hover_function_test.rs:117:9:
assertion failed: ws.check_hover(r#"
                ---@type Hover.Test3<string>
                local test3

                local <??>event = test3.event
            "#,
    VirtualHoverResult {
        value: "```lua\n(method) Test3:event(event: \"B\", key: string)\n```\n\n&nbsp;&nbsp;in class `Hover.Test3`\n\n---\n\n---\n\n```lua\n(method) Test3:event(event: \"A\", key: string)\n```".to_string(),
    })
```

PR #647, opened 2025-07-17
Fix doc tag associations, warn about useless tags

### LINKED ISSUE #575: Support a generic declaration for holding type metadata
In Recoil we're using emmylua cli to generate our documentation and lua library from source.

We have a very complex scenario where multiple lua environments are present, with different types available on each.

Not only that but we also want to be able to customize the way our documentation is rendered further than current emmylua_doc_cli markdown export capabilities, for example grouping certain declarations together.

You can find the output of our current build [here](https://beyond-all-reason.github.io/RecoilEngine/docs/lua-api/globals/spring/#springcreategrounddecal) to have some idea of the challenge, notice how it's very difficult to explore such a large API space and find what's your use case scenario.

Proposal is a declaration that can hold some textual value that is relayed on to the final json export, then we can handle the json via a script to generate our documentation pages as we want, example:

```lua
--- @function Spring.CreateGroundDecal
--- @tag section:ground_decals type:write
 
--- @function Spring.GetAllGroundDecals
--- @tag section:ground_decals type:read
```

Of course, such a declaration can be used in the markdown export in the future, but right now we only really this data to be available on the json export data.

Then we should be able to group declarations together on our documentation on sections, if they write or get state, etc...

If one provides some tips on how to achieve this I could try taking a stab at it.

taminomara:
This commit fixes doc tag association for tags that appear after `@class` and other decls. It affects `@see`, `@return_cast`, `@source`, `@version`, `@deprecated`, visibility tags and custom (other) tags.

It also adds diagnostics for doc tags that couldn't be attached to any object, and therefore are ignored.

Fix #575.

### DIFF
--- a/crates/emmylua_code_analysis/src/compilation/analyzer/doc/tags.rs
+++ b/crates/emmylua_code_analysis/src/compilation/analyzer/doc/tags.rs
@@ -4,7 +4,7 @@ use emmylua_parser::{
 
 use crate::{
     db_index::{LuaMemberId, LuaSemanticDeclId, LuaSignatureId},
-    LuaDeclId,
+    AnalyzeError, DiagnosticCode, LuaDeclId,
 };
 
 use super::{
@@ -74,8 +74,8 @@ pub fn analyze_tag(analyzer: &mut DocAnalyzer, tag: LuaDocTag) -> Option<()> {
         LuaDocTag::Version(version) => {
             analyze_version(analyzer, version)?;
         }
-        LuaDocTag::Async(_) => {
-            analyze_async(analyzer)?;
+        LuaDocTag::Async(tag) => {
+            analyze_async(analyzer, tag)?;
         }
 
         // field or operator
@@ -135,7 +135,23 @@ pub fn find_owner_closure(analyzer: &DocAnalyzer) -> Option<LuaClosureExpr> {
     None
 }
 
+pub fn find_owner_closure_or_report(
+    analyzer: &mut DocAnalyzer,
+    tag: &impl LuaAstNode,
+) -> Option<LuaClosureExpr> {
+    match find_owner_closure(analyzer) {
+        Some(id) => Some(id),
+        None => {
+            report_orphan_tag(analyzer, tag);
+            None
+        }
+    }
+}
+
 pub fn get_owner_id(analyzer: &mut DocAnalyzer) -> Option<LuaSemanticDeclId> {
+    if let Some(current_type_id) = &analyzer.current_type_id {
+        return Some(LuaSemanticDeclId::TypeDecl(current_type_id.clone()));
+    }
     let owner = analyzer.comment.get_owner()?;
     match owner {
         LuaAst::LuaAssignStat(assign) => {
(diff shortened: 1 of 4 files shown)


crates/emmylua_code_analysis/src/compilation/analyzer/doc/tags.rs
```diff
@@ -135,7 +135,23 @@ pub fn find_owner_closure(analyzer: &DocAnalyzer) -> Option<LuaClosureExpr> {
...
+        Some(id) => Some(id),
+        None => {
+            report_orphan_tag(analyzer, tag);
+            None
+        }
+    }
+}
+
 pub fn get_owner_id(analyzer: &mut DocAnalyzer) -> Option<LuaSemanticDeclId> {
+    if let Some(current_type_id) = &analyzer.current_type_id {
```
taminomara, line 152: Note: this is the main change, the rest is just error reporting.

badosu:
Thank you so much!

lewis6991:
In Neovim we use other custom tags (e.g. `@nodoc`) for documentation generation.

We disable `unknown-doc-tag` to prevent warnings, but with this change we also now need to disable `annotation-usage-error`.

I think `annotation-usage-error` should only apply to annotations supported by emmylua.

clason:
Sorry to do that again, but this PR introduced a regression (on nvim-treesitter, main branch):
```
--- nvim-treesitter/parsers.lua [1 error]
error: `@type nvim-ts.parsers` can't be used here [annotation-usage-error]
  -->: nvim-treesitter/parsers.lua:1:5

  1 │ ---@type nvim-ts.parsers
  2 │ return {
 ```

`nvim-ts.parsers` is defined as
```lua
---@alias nvim-ts.parsers table<string,ParserInfo>
```

taminomara:
> Sorry to do that again, but this PR introduced a regression (on nvim-treesitter, main branch):

This is not a regression, `@type` has no effect when attached to a `return` statement.

taminomara:
> I think annotation-usage-error should only apply to annotations supported by emmylua.

This warning happens when a tag has no effect, i.e. it's not attached to anything. Unless you roll a custom source parser, there's really no point in disabling these warnings. Can you elaborate on how you use custom tags?

clason:
> This is not a regression, @type has no effect when attached to a return statement.

The regression is that it now _does_: it throws an error where it didn't before (and doesn't anymore now).

clason:
> This warning happens when a tag has no effect

Then it should be a hint (like unused variable or unnecessary assert), not a hard error!

taminomara:
Maybe we should demote `annotation-usage-error` to a warning then? Can you start make a separate issue?

Meanwhile, you can set a custom level for `annotation-usage-error` in `.emmylua.rc`:

```json
{
  "diagnostics": {
    "severity": {
      "annotation-usage-error": "warning"
    }
  }
}
```

For context, right now, there's only one additional case when we emit `annotation-usage-error`: if you use `@field` before `@class`.

clason:
(My issue is already fixed, so I'm not going to open an issue about it.)

taminomara:
> We disable unknown-doc-tag to prevent warnings, but with this change we also now need to disable annotation-usage-error.

BTW you can add custom tags into `.emmylua.rc` instead:

```json
{
  "doc": {
    "knownTags": ["foo", ...]
  }
}
```

lewis6991:
> Unless you roll a custom source parser,

We have a custom parser, rolled with lpeg.

We currently use the tags `@nodoc`, `@inlinedoc` and `@brief` to aid with docgen.

lewis6991:
> This is not a regression, @type has no effect when attached to a return statement.

We use this to coerce the type of the expression in the return statement. Afaik this works in LuaLS.

It should be semantically equivalent to.

```lua
--- @type TYPE
local ret = EXPR

return ret
```

xuhuanzy:
> This is not a regression, @type has no effect when attached to a return statement.

The behavior is incorrect; `---@type` should be allowed to be attached to a return statement.

taminomara:
> The behavior is incorrect; ---@type should be allowed to be attached to a return statement.

As of right now, it doesn't have any effect.

PR #665, opened 2025-07-25
Support highlighting Markdown/RST in comments

taminomara:
This PR adds an optional feature that highlights Markdown, MySt (a Markdown extension used with Sphinx), and RST syntax in comments and provides Go To Definition implementation for cross-references. This feature is disabled by default; it can be enabled by specifying `doc.syntax` in `.emmyrc.json`.

image

# Caveats

- We use Semantic Tokens to provide syntax highlighting. Since there are no Semantic Tokens for bold and italic text, those are not highlighted.

### DIFF
--- a/crates/emmylua_code_analysis/src/semantic/mod.rs
+++ b/crates/emmylua_code_analysis/src/semantic/mod.rs
@@ -40,7 +40,7 @@ pub use visibility::check_export_visibility;
 use visibility::check_visibility;
 
 use crate::semantic::member::find_members_with_key;
-use crate::{Emmyrc, LuaDocument, LuaSemanticDeclId, db_index::LuaTypeDeclId};
+use crate::{Emmyrc, LuaDocument, LuaSemanticDeclId, ModuleInfo, db_index::LuaTypeDeclId};
 use crate::{
     FileId,
     db_index::{DbIndex, LuaType},
@@ -85,6 +85,10 @@ impl<'a> SemanticModel<'a> {
 
     pub fn get_document(&self) -> LuaDocument {
         self.db.get_vfs().get_document(&self.file_id).unwrap()
+    }
+
+    pub fn get_module(&self) -> Option<&ModuleInfo> {
+        self.db.get_module_index().get_module(self.file_id)
     }
 
     pub fn get_document_by_file_id(&self, file_id: FileId) -> Option<LuaDocument> {
@@ -254,6 +258,10 @@ impl<'a> SemanticModel<'a> {
 
     pub fn get_emmyrc(&self) -> &Emmyrc {
         &self.emmyrc
+    }
+
+    pub fn get_emmyrc_arc(&self) -> Arc<Emmyrc> {
+        self.emmyrc.clone()
     }
 
     pub fn get_root(&self) -> &LuaChunk {
--- a/crates/emmylua_ls/src/handlers/test_lib/mod.rs
+++ b/crates/emmylua_ls/src/handlers/test_lib/mod.rs
@@ -2,10 +2,11 @@ use emmylua_code_analysis::{EmmyLuaAnalysis, Emmyrc, FileId, VirtualUrlGenerator
 use googletest::prelude::*;
 use itertools::Itertools;
 use lsp_types::{
-    CodeActionOrCommand, CompletionItem, CompletionItemKind, CompletionResponse,
-    CompletionTriggerKind, GotoDefinitionResponse, Hover, HoverContents, InlayHintLabel, Location,
-    MarkupContent, Position, SemanticTokenModifier, SemanticTokenType, SemanticTokensResult,
-    SignatureHelpContext, SignatureHelpTriggerKind, SignatureInformation, TextEdit,
+    ClientCapabilities, CodeActionOrCommand, CompletionItem, CompletionItemKind,
+    CompletionResponse, CompletionTriggerKind, GotoDefinitionResponse, Hover, HoverContents,
+    InlayHintLabel, Location, MarkupContent, Position, SemanticTokenModifier, SemanticTokenType,
+    SemanticTokensResult, SignatureHelpContext, SignatureHelpTriggerKind, SignatureInformation,
+    TextEdit,
 };
 use std::collections::HashSet;
 use std::{ops::Deref, sync::Arc};
@@ -489,9 +490,14 @@ impl ProviderVirtualWorkspace {
         expected: Vec<VirtualSemanticToken>,
     ) -> Result<()> {
         let file_id = self.def(block_str);
-        let result = semantic_token(&self.analysis, file_id, ClientId::VSCode)
-            .ok_or("failed to get semantic tokens")
-            .or_fail()?;
+        let result = semantic_token(
+            &self.analysis,
+            file_id,
+            &ClientCapabilities::default(),
+            ClientId::VSCode,
+        )
+        .ok_or("failed to get semantic tokens")
+        .or_fail()?;
         let SemanticTokensResult::Tokens(result) = result else {
             return fail!("expected SemanticTokensResult::Tokens, got {result:?}");
         };
--- a/crates/emmylua_ls/src/handlers/test_lib/mod.rs
+++ b/crates/emmylua_ls/src/handlers/test_lib/mod.rs
@@ -214,7 +214,7 @@ impl ProviderVirtualWorkspace {
     pub fn check_completion_with_kind(
         &mut self,
         block_str: &str,
-        expected: Vec<VirtualCompletionItem>,
+        mut expected: Vec<VirtualCompletionItem>,
         trigger_kind: CompletionTriggerKind,
     ) -> Result<()> {
         let (content, position) = Self::handle_file_content(block_str)?;
@@ -229,10 +229,13 @@ impl ProviderVirtualWorkspace {
         .ok_or("failed to get completion")
         .or_fail()?;
         // 对比
-        let items = match result {
+        let mut items = match result {
             CompletionResponse::Array(items) => items,
             CompletionResponse::List(list) => list.items,
         };
+
+        items.sort_by_key(|item| item.label.clone());
+        expected.sort_by_key(|item| item.label.clone());
 
         fn get_item_detail(i: &CompletionItem) -> Option<&String> {
             i.label_details.as_ref().and_then(|d| d.detail.as_ref())
--- a/crates/emmylua_parser/src/lexer/lua_lexer.rs
+++ b/crates/emmylua_parser/src/lexer/lua_lexer.rs
@@ -693,7 +693,7 @@ impl<'a> LuaLexer<'a> {
 
         if self.reader.current_char().is_alphabetic() {
             let ch = self.reader.current_char();
-            self.error(|| format!("unexpected character '{ch}' after number literal"));
+            self.error(|| t!("unexpected character '%{ch}' after number literal", ch = ch));
         }
 
         match state {
--- a/crates/emmylua_parser/src/lexer/lua_lexer.rs
+++ b/crates/emmylua_parser/src/lexer/lua_lexer.rs
@@ -30,7 +30,7 @@ impl LuaLexer<'_> {
                 break;
             }
 
-            tokens.push(LuaTokenData::new(kind, self.reader.saved_range()));
+            tokens.push(LuaTokenData::new(kind, self.reader.current_range()));
         }
 
         tokens
@@ -122,7 +122,7 @@ impl LuaLexer<'_> {
                 if self.reader.current_char() != '[' {
                     self.errors.push(LuaParseError::syntax_error_from(
                         &t!("invalid long string delimiter"),
-                        self.reader.saved_range(),
+                        self.reader.current_range(),
                     ));
                     return LuaTokenKind::TkLongString;
                 }
@@ -149,7 +149,7 @@ impl LuaLexer<'_> {
                         if !self.lexer_config.support_integer_operation() {
                             self.errors.push(LuaParseError::syntax_error_from(
                                 &t!("bitwise operation is not supported"),
-                                self.reader.saved_range(),
+                                self.reader.current_range(),
                             ));
                         }
 
@@ -176,7 +176,7 @@ impl LuaLexer<'_> {
                         if !self.lexer_config.support_integer_operation() {
                             self.errors.push(LuaParseError::syntax_error_from(
                                 &t!("bitwise operation is not supported"),
-                                self.reader.saved_range(),
+                                self.reader.current_range(),
                             ));
                         }
 
@@ -198,7 +198,7 @@ impl LuaLexer<'_> {
                     if !self.lexer_config.support_integer_operation() {
                         self.errors.push(LuaParseError::syntax_error_from(
                             &t!("bitwise operation is not supported"),
-                            self.reader.saved_range(),
+                            self.reader.current_range(),
                         ));
                     }
                     return LuaTokenKind::TkBitXor;
@@ -252,7 +252,7 @@ impl LuaLexer<'_> {
                 if self.reader.current_char() != quote {
                     self.errors.push(LuaParseError::syntax_error_from(
                         &t!("unfinished string"),
-                        self.reader.saved_range(),
+                        self.reader.current_range(),
                     ));
                     return LuaTokenKind::TkString;
                 }
@@ -297,7 +297,7 @@ impl LuaLexer<'_> {
                                 _ if self.reader.is_eof() => {
                                     self.errors.push(LuaParseError::syntax_error_from(
                                         &t!("unfinished long comment"),
-                                        self.reader.saved_range(),
+                                        self.reader.current_range(),
                                     ));
                                     return LuaTokenKind::TkLongComment;
                                 }
@@ -323,7 +323,7 @@ impl LuaLexer<'_> {
                         if !self.lexer_config.support_integer_operation() {
                             self.errors.push(LuaParseError::syntax_error_from(
                                 &t!("integer division is not supported"),
-                                self.reader.saved_range(),
+                                self.reader.current_range(),
                             ));
                         }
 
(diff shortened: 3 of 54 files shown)


taminomara:
Note: this is a rather large PR, see individual commits to get more digestible diffs.

CppCXY:
Sorry, but I do not agree with any changes to the AST, especially those that use regular expressions to match and modify the usage of regions.

taminomara:
> especially those that use regular expressions to match and modify the usage of regions.

I can redo those without regexes, but might you tell me what exactly you disagree with and why?

- requiring spaces after region markers,
- adding new tokens,
- adding new nodes,
- the whole concept in general?

taminomara:
...
- handling of whitespaces in beginning of comments?

CppCXY:
You just want to highlight the documentation in comments, so you only need to parse the comments once during semantic token request and highlight them in the appropriate places, rather than modifying the entire structure of the syntax tree.

taminomara:
Hm, I see your point. I can implement this. Although there will be complications if we want to support Go To Definition in comments.

I see it like this:

- make a cache for comment markup, in case if IDE requests semantic tokens twice;
- in Go To Definition request, if the user clicked inside of the documentation comment, we can check the comment markup cache to see if there's a cross reference in that place;
- in Document Selection Range request, we can either ignore comments, or use comment markup cache once again.

Would you agree with this implementation?

CppCXY:
This is a multithreaded language server, and caching introduces data races

CppCXY:
Even if you re-parse everything, it will still be faster than using a cache.

taminomara:
Comment markup cache doesn't need its own AST, it will only save highlighted ranges.

taminomara:
> caching introduces data races

I can make cache async/multi-threaded, that's not a big issue.

> Even if you re-parse everything, it will still be faster than using a cache.

That's possible, I'll need to benchmark this.

taminomara:
I'll start with re-parsing on every request, and see if caching is warranted later.

Two more questions:

1. do you agree with changes in 06cc9824? I've struggled with this behavior quite a bit, because no other language server does this.
2. would you be interested in a separate PR that requires a space/end-of-line after "#region", provided I don't use regexes for that? Current behavior seems a bit too unspecific to me.

taminomara:
Reworked this PR to avoid changes to AST.

Caching wasn't required, it's fast enough to re-parse docs on every relevant request.

crates/emmylua_parser/src/desc_parser/md.rs
```diff
@@ -0,0 +1,1757 @@
...
+        reader: &mut Reader<'a>,
+    ) -> Result<(), ()> {
+        let bt = BacktrackPoint::new(self, reader);
+
+        reader.eat_while(is_ws);
+        if !reader.tail_text().starts_with("---") {
+            bt.rollback(self, reader);
+            return Err(());
+        }
+        if !is_blank(&reader.tail_text()[3..]) {
```
CppCXY: normal description will only 2 char, Out-of-bounds access may occur here.

crates/emmylua_code_analysis/src/semantic/mod.rs
```diff
@@ -87,6 +87,10 @@ impl<'a> SemanticModel<'a> {
         self.db.get_vfs().get_document(&self.file_id).unwrap()
     }
 
+    pub fn get_module(&self) -> &ModuleInfo {
```
CppCXY: Use Option<&ModuleInfo>, If it is meta and marked with no-require or _, its module_info may not exist.

crates/emmylua_parser/src/desc_parser/md.rs
CppCXY: normal description will only 2 char, Out-of-bounds access may occur here.

crates/emmylua_parser/src/text/reader.rs
```diff
@@ -21,68 +54,89 @@ impl<'a> Reader<'a> {
...
-        if !self.start {
-            self.start = true;
-            self.bump();
+        self.current_buffer_byte_pos += self.current_buffer_byte_len;
+        self.current_buffer_byte_len = 0;
+    }
+
+    pub fn reset_buff_into_sub_reader(&mut self) -> Reader<'a> {
+        let mut reader = Reader::new_with_range(self.current_text(), self.current_range());
+        if let Some(prev) = &self.text[..self.current_buffer_byte_pos]
```
CppCXY, line 91: Is this implementation too resource-intensive?

crates/emmylua_parser/src/text/reader.rs
```diff
@@ -115,12 +181,22 @@ impl<'a> Reader<'a> {
...
         self.text
     }
 
     pub fn get_current_end_pos(&self) -> usize {
-        self.save_buffer_byte_pos + self.save_buffer_byte_len
+        self.current_buffer_byte_pos + self.current_buffer_byte_len
+    }
+}
+
+impl<'a> Default for Reader<'a> {
```
CppCXY: What is the purpose of implementing default for it?

CppCXY:
When the code block area uses Lua syntax, should we use luaLexer to parse it and highlight keywords, strings, and numbers?

crates/emmylua_parser/src/desc_parser/md.rs
taminomara: This section deals with MySt syntax for directive parameters:

````md
```{code-block} lua
---
linenos: true
emphasize-lines: [1, 3]
---
function example()
    ...
end
```
````

out of bounds access is prevented by the condition above:

```rust
if !reader.tail_text().starts_with("---")
```

crates/emmylua_parser/src/text/reader.rs
taminomara: I've checked, `chars()` is a bi-directional iterator, `next_back()` is `O(1)`.

taminomara:
> When the code block area uses Lua syntax, should we use luaLexer to parse it and highlight keywords, strings, and numbers?

Done:

image

CppCXY:
Your changes to error have caused the error location to change.
image


```lua
local p

--- @return fun(): (integer?, any?, ...)
local function f(tasks)
end
```

CppCXY:
I got error:
```log

thread 'tokio-runtime-worker' panicked at crates\emmylua_parser\src\text\text_range.rs:18:9:
assertion failed: start_offset <= end_offset
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

thread 'tokio-runtime-worker' panicked at crates\emmylua_parser\src\text\text_range.rs:18:9:
assertion failed: start_offset <= end_offset

```

```lua
---```lua
---
--- local t = 213
---```
---
--- .. code-block:: lua
---
---    local t = 123
---    yes = 1123
local t = 123

```

taminomara:
Fixed bugs, added more tests, implemented autocomplete for ref targets, improved go to definition to handle modules.

crates/emmylua_parser/src/lexer/lua_lexer.rs
```diff
@@ -523,18 +552,26 @@ impl LuaLexer<'_> {
...
         if self.reader.current_char().is_alphabetic() {
-            self.errors.push(LuaParseError::syntax_error_from(
-                &format!(
-                    "unexpected character '{}' after number literal",
-                    self.reader.current_char()
-                ),
-                self.reader.saved_range(),
-            ));
+            let ch = self.reader.current_char();
+            self.error(|| format!("unexpected character '{ch}' after number literal"));
```
CppCXY: use t!

crates/emmylua_parser/src/lexer/lua_lexer.rs
```diff
@@ -523,18 +552,26 @@ impl LuaLexer<'_> {
...
         }
 
         match state {
             NumberState::Int | NumberState::Hex => LuaTokenKind::TkInt,
             _ => LuaTokenKind::TkFloat,
         }
     }
+
+    fn error<F, R>(&mut self, msg: F)
```
CppCXY, line 705: Why not add an `fn error_msg`?

crates/emmylua_parser/src/desc_parser/dummy.rs
```diff
@@ -0,0 +1,10 @@
+use crate::LuaDocDescription;
+use crate::desc_parser::{DescItem, LuaDescParser};
+
+pub struct DummyParser;
+
+impl LuaDescParser for DummyParser {
```
CppCXY: What is the purpose of this?

CppCXY:
image
Why does the highlighting look like this?

CppCXY:
If it works well, I plan to enable Markdown mode by default. This is how it currently performs in nvim. 
image

What’s your opinion? @lewis6991

lewis6991:
To begin with, I really don't think this should be enabled by default. I will use this feature, but based on conversations I've had with people, this will be polarising to a lot of people, especially if there are any issues, even minor ones. I suggest this is opt-in until we know it is stable.

- Does it only parse comments with `---` or all comments? If not I suggest this is limited to only `---` comments.

- Can we configure whether to use markdown or rst?

- This is a huge PR. Would it make sense to place the markdown and rst parser into separate crates?

CppCXY:
> * Does it only parse comments with `---` or all comments? If not I suggest this is limited to only `---` comments.

I didn’t distinguish between --- and --, but it should be possible to check with starts_with.

> * Can we configure whether to use markdown or rst?

currently you can config:
```json
    "doc": {
        "syntax": "md" -- none, md, myst, rst
    }
```

> * This is a huge PR. Would it make sense to place the markdown and rst parser into separate crates?

I think so.

taminomara:
> Why does the highlighting look like this?

`.. code-block::` is an RST syntax, markdown parser sees it as simple text. After processing it, the parser sees something indented by 4 spaces, and parses it as a generic code block with no specified syntax.

So, this is how this code is seen by parser:

````md
```lua
Lua code.
```

Text.

    Simple code.
````

As you can see, GitHub syntax highlighting does exactly the same thing:

````md
```lua
 local t = 213
```

.. code-block:: lua

    Code indented by 4 spaces.
````

---

As to `local` keyword not being highlighted in VSCode, I'll fix that.

File: crates/emmylua_parser/src/lexer/lua_lexer.rs (hunk 6 above)
Comment (taminomara) [reply crates/emmylua_parser/src/lexer/lua_lexer.rs:705]: Not sure what you mean.

If it's about accepting a closure, it's just a micro optimization so that we don't run translation if error reporting is disabled.

**Comment by taminomara** [comment]
> based on conversations I've had with people, this will be polarising to a lot of people

Just for context, is that because highlighting in comments is distracting for some people, or are there other concerns?

> Does it only parse comments with --- or all comments? If not I suggest this is limited to only --- comments.

I've disabled highlighting for any comment that doesn't start with three dashes. It should also detect comment blocks adorned with long dashed lines:

```lua
---------------   <- not highlighted
--- Comment       <- highlighted
---------------   <- highlighted
--- Comment       <- highlighted
---------------   <- not highlighted

---------------   <- not highlighted
-- Comment        <- not highlighted
---------------   <- not highlighted
-- Comment 2      <- not highlighted
---------------   <- not highlighted
```

> Can we configure whether to use markdown or rst?

Yes. Also, `std` will use markdown even if user configured rst.

lewis6991:
> Just for context, is that because highlighting in comments is distracting for some people, or are there other concerns?

Yes, and I've also had someone say they want treesitter (which has more highlights and is faster) to handle this kind of highlighting, not the server.

taminomara:
> Yes, and I've also had someone say they want treesitter (which has more highlights and is faster) to handle this kind of highlighting, not the server.

Then there should be a way to disable highlighting even if `emmyrc` specifies otherwise. I think I'll do it as a separate PR though.

taminomara:
I ran parsing and completions on my codebase, fixed some minor things. I believe I've done all features I wanted for this PR, it's ready for merge.

In next PR I'll add option to force-disable markdown parsing even if `.emmyrc` enables it; it will be coupled with updates to VSCode plugin to expose this new option to users.

clason:
Please don't invent new off-spec options that only VS Code is set up to handle it. It's enough to make this opt-in via `.emmrc.json` (or the LSP configuration requests), anything beyond that makes it much more (and too) difficult to reason about.

If you want to do anything, split the `doc.syntax` option into two: one for enabling/disabling it, and one that controls the type of syntax. This way, the latter can be set per-configuration in a local `.emmyrc.json` while enabling this extension can be done in the client configuration.

I would also (strongly) recommend to name and structure the options in a way to make clear that they relate to semantic tokens. You probably also want to document the (off-spec?) tokens you send for documentation.

taminomara:
> If you want to do anything, split the doc.syntax option into two

Do you think it's needed? I like this idea, but I don't have a strong opinion here.

> I would also (strongly) recommend to name and structure the options in a way to make clear that they relate to semantic tokens.

It's not just semantic tokens, though, but also hovers, completion, text ranges, and go to definition. Plus, option names must reflect that they affect highlighting markup within descriptions, and don't affect highlighting `@tags`, types, and so on.

So, this is what I propose then (ideas for better naming are welcome):

- `doc.syntax` for overall syntax configuration,
- `semantic_tokens.enableForDocumentationMarkup` to enable/disable semantic tokens,
- not sure if options for other sub-systems are necessary, but we can have `hover.enableForDocumentationMarkup` and others.

As for defaults, I want to avoid scenario where a user sets `doc.syntax` and doesn't see anything because `semantic_tokens.enableForDocumentationMarkup` is disabled. So we can either have `doc.syntax = "none"` and `semantic_tokens.enableForDocumentationMarkup = true`, or `doc.syntax = "md"` and `semantic_tokens.enableForDocumentationMarkup = false`. IMO, the latter option seems more reasonable.

> Please don't invent new off-spec options that only VS Code is set up to handle it.

With settings being more granular this is indeed not necessary. Bear in mind, though, that there's already lots of off-spec things happening in VSCode extension, mostly because (1) VSCode doesn't support overlapping semantic tokens, and (2) VSCode doesn't allow customizing LSP configuration call the same way neovim does. So some changes will have to be made regardless.

> You probably also want to document the (off-spec?) tokens you send for documentation.

Actually, no off-spec tokens so far. Maybe I'll add some for italic/bold text, but that will be later).

clason:
I don't think the "active" LSP features need to be opt-in, as they need to be actively triggered on a documentation location anyway. (It would be different if references also included documentation locations -- which I feel they shouldn't either way.)

I still -- strongly -- recommend the option split, since the syntax format is a project choice while the highlighting is a user choice. This needs to be respected. If this requires making doc semantic tokens opt-out, so be it. 

And I understand that the extension needs special support for this; that is fine. I amjust concerned with the server configuration not being driven by that -- everything should be (also) configurable through standard interfaces (config file and ConfigurationChanged notifications).

taminomara:
> I don't think the "active" LSP features need to be opt-in

There's also selection ranges, which is not an active feature. If we're worrying about stability, this should also be opt-in, at least for now.

Also, hover is not an "active" feature, but it's easy to trigger it accidentally. Should it be opt-in as well?

taminomara:
Added `semanticTokens.renderDocumentationMarkup`, default is `false`.

The reason for this name is for consistency with possible future options.

For example, currently we don't provide `documentation` modifier for semantic tokens related to documentation comments. In future, we might implement this, but we'll need to make it opt-out because at least in VSCode, most themes render documentation comments in horrendously bright colors. Option name `renderDocumentationAsComments` will sound consistent with `renderDocumentationMarkup`.

> And I understand that the extension needs special support for this; that is fine. I amjust concerned with the server configuration not being driven by that -- everything should be (also) configurable through standard interfaces (config file and ConfigurationChanged notifications).

I've created #680 to address this issue. When it's merged, VSCode and NeoVim (and other editors) will be able to configure things using the same mechanism.

crates/emmylua_parser_desc/src/testlib.rs
```diff
@@ -0,0 +1,70 @@
...
+        .get_red_root()
+        .descendants()
+        .filter(|node| matches!(node.kind(), LuaKind::Syntax(LuaSyntaxKind::DocDescription)))
+        .next()
+    else {
+        panic!("No desc found in {:?}", tree.get_red_root());
+    };
+    let ranges = parser.parse(code, LuaDocDescription::cast(desc).unwrap());
+    let result = format_result(code, ranges);
+    assert_eq!(result.trim(), expected.trim());
```
CppCXY, line 19: If assert is written inside the function, will the error location always be inside this function when an error occurs?

crates/emmylua_ls/src/handlers/test_lib/mod.rs
```diff
@@ -195,29 +191,24 @@ impl ProviderVirtualWorkspace {
             CancellationToken::new(),
         );
         let Some(result) = result else {
-            return false;
+            panic!("completion result is empty");
```
CppCXY: If panic occurs inside the function, does that mean I can’t quickly jump to the exact error location when an error happens?

crates/emmylua_parser_desc/src/util.rs
```diff
@@ -0,0 +1,504 @@
...
+    }
+}
+
+pub fn is_opening_quote(c: char) -> bool {
+    if c.is_ascii() {
+        matches!(c, '\'' | '"' | '<' | '(' | '[' | '{')
+    } else {
+        static RE: LazyLock<regex::Regex> =
+            LazyLock::new(|| regex::Regex::new(r"^\p{Ps}|\p{Pi}|\p{Pf}$").unwrap());
+        let mut tmp = [0; 4];
```
CppCXY: I don’t want to introduce regular expressions into the base library. Maybe you can use unicode-general-category instead.

crates/emmylua_parser_desc/src/testlib.rs
taminomara: RustRover highlights relevant lines in test code, but it seems that VSCode doesn't. I'll change this function to return `Result` with an error message, this will be more ergonomic than returning a boolean.

CppCXY:
please donot change my `local` keyword color
now: 
image

before:
image

CppCXY:
please resolve conflicts

CppCXY:
If everything works well, I will merge it before the release tomorrow, but please resolve the conflicts first.

taminomara:
Rebased.

Since you want to release tomorrow, I've also added support for bold/italic highlighting via emmy annotator, see also EmmyLua/VSCode-EmmyLua#248.

PR #669, opened 2025-07-28
refactor member analysis

CppCXY:
Previously, member analysis was highly dynamic. We needed to analyze the parent class and add members at the same time, and when accessing a class member, we couldn’t assume that the field didn’t exist, so we repeatedly tried to resolve it during the resolve phase. I think we should more statically determine which members a class has, and we will implement this based on some more aggressive assumptions.


CppCXY:
The work did not meet expectations.

PR #680, opened 2025-08-03
Unify initialization process for all editors

taminomara:
Right now, our LSP only requests settings when client is VSCode or NeoVim. To make matters worse, it requests *different* settings for these clients:

- for VSCode, it requests VSCode-specific about files and extensions,
- for NeoVim, it requests settings in scope `"Lua"`, and uses them as overrides for `.emmyrc`.

This PR aims to unify this behavior and make it easier for users to override settings without having to create `.emmyrc.json`. This is especially handy when you want to override a setting for a single project, but don't want to change project's `.emmyrc.json`. For example, switching documentation highlighting on per-project basis will become easier (see discussion in #665).

# Changes

- When client is VSCode, LSP will request settings under `"emmylua"` scope, as well as VSCode-specific settings mentioned above.

  I'll add relevant settings to the plugin settings page in a separate PR when this is merged; until then, these changes will not affect user experience.

- When client is NeoVim, LSP will request settings under `"emmylua"` scope. If there's no such scope, it will fall back to using the `"Lua"` scope.

  Currently, both LuaLs and EmmyLua use the same scope `"Lua"`. It works in NeoVim because options for every LSP server are independent. In VSCode, however, they all share the same namespace, meaning that we can't reuse `"Lua"`. I'd like to keep things consistent, hence this rename.

  I'll send a PR to update nvim-lspconfig's documentation for EmmyLua when this is merged.

- For any other client, LSP will request settings under `"emmylua"` scope.

  Currently, there are no clients that provide this scope (EmmyLua2 for Intellij doesn't provide any config scopes), so it shouldn't break anything, but it will provide future client implementations with a better way to override LSP configuration.

# Tests

Tested this with VSCode and NeoVim.

### DIFF
--- a/crates/emmylua_ls/src/handlers/initialized/client_config/mod.rs
+++ b/crates/emmylua_ls/src/handlers/initialized/client_config/mod.rs
@@ -20,6 +20,7 @@ pub struct ClientConfig {
 pub async fn get_client_config(
     context: &ServerContextSnapshot,
     client_id: ClientId,
+    supports_config_request: bool,
 ) -> ClientConfig {
     let mut config = ClientConfig {
         client_id,
@@ -29,11 +30,16 @@ pub async fn get_client_config(
         partial_emmyrcs: None,
     };
     match client_id {
-        ClientId::VSCode => get_client_config_vscode(context, &mut config).await,
-        ClientId::Neovim => {
-            get_client_config_default(context, &mut config, Some(&["Lua", "emmylua"])).await
+        ClientId::VSCode => {
+            get_client_config_vscode(context, &mut config).await;
         }
-        _ => get_client_config_default(context, &mut config, None).await,
+        ClientId::Neovim => {
+            get_client_config_default(context, &mut config, Some(&["Lua", "emmylua"])).await;
+        }
+        _ if supports_config_request => {
+            get_client_config_default(context, &mut config, None).await;
+        }
+        _ => {}
     };
 
     config
--- a/crates/emmylua_ls/src/context/mod.rs
+++ b/crates/emmylua_ls/src/context/mod.rs
@@ -10,6 +10,7 @@ pub use client_id::{ClientId, get_client_id};
 use emmylua_code_analysis::EmmyLuaAnalysis;
 pub use file_diagnostic::FileDiagnostic;
 use lsp_server::{Connection, ErrorCode, Message, RequestId, Response};
+use lsp_types::ClientCapabilities;
 pub use snapshot::ServerContextSnapshot;
 pub use status_bar::ProgressTask;
 pub use status_bar::StatusBar;
@@ -29,10 +30,11 @@ pub struct ServerContext {
     file_diagnostic: Arc<FileDiagnostic>,
     workspace_manager: Arc<RwLock<WorkspaceManager>>,
     status_bar: Arc<StatusBar>,
+    client_capabilities: Arc<ClientCapabilities>,
 }
 
 impl ServerContext {
-    pub fn new(conn: Connection) -> Self {
+    pub fn new(conn: Connection, client_capabilities: Arc<ClientCapabilities>) -> Self {
         let client = Arc::new(ClientProxy::new(Connection {
             sender: conn.sender.clone(),
             receiver: conn.receiver.clone(),
@@ -60,6 +62,7 @@ impl ServerContext {
             cancllations: Arc::new(Mutex::new(HashMap::new())),
             workspace_manager,
             status_bar,
+            client_capabilities,
         }
     }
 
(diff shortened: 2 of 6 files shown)


crates/emmylua_ls/src/handlers/initialized/client_config/default_config.rs
```diff
@@ -0,0 +1,63 @@
...
+) -> Option<()> {
+    let workspace_folders = context
+        .workspace_manager
+        .read()
+        .await
+        .workspace_folders
+        .clone();
+    let main_workspace_folder = workspace_folders.get(0);
+    let client = &context.client;
+    let scope_uri = main_workspace_folder.map(|p| file_path_to_uri(p).unwrap());
```
taminomara, line 24: Note: this rolls back a change [made in previous commit](https://github.com/EmmyLuaLs/emmylua-analyzer-rust/commit/9464063492bd577b8fdbf9f985bde64928278aee#diff-a24411316d4e6cbe920f2b6f662b7d2a5e84c24077749245a2133a534da4f95fR22-R24). This is because NeoVim doesn't require workspace root to run EmmyLua. In this case, `workspace_folders` will be empty, but we should still request configs from the client.

Let me know if I've missed something and I shouldn't change this.

crates/emmylua_ls/src/handlers/initialized/client_config/mod.rs
```diff
@@ -30,8 +30,10 @@ pub async fn get_client_config(
     };
     match client_id {
         ClientId::VSCode => get_client_config_vscode(context, &mut config).await,
-        ClientId::Neovim => get_client_config_neovim(context, &mut config).await,
-        _ => Some(()),
+        ClientId::Neovim => {
+            get_client_config_default(context, &mut config, Some(&["Lua", "emmylua"])).await
+        }
+        _ => get_client_config_default(context, &mut config, None).await,
```
CppCXY: Not all clients support workspace/configuration, so sending it to other clients may cause the language server to freeze for several seconds.

crates/emmylua_ls/src/handlers/initialized/client_config/mod.rs
```diff
@@ -30,8 +30,10 @@ pub async fn get_client_config(
     };
     match client_id {
         ClientId::VSCode => get_client_config_vscode(context, &mut config).await,
-        ClientId::Neovim => get_client_config_neovim(context, &mut config).await,
-        _ => Some(()),
+        ClientId::Neovim => {
+            get_client_config_default(context, &mut config, Some(&["Lua", "emmylua"])).await
```
CppCXY: I think since LuaLS has already become the user preference for Neovim, there’s no need to request emmyLua anymore. Also, “Lua” is capitalized, while “emmylua” is lowercase, which doesn’t look as harmonious.

crates/emmylua_ls/src/handlers/initialized/client_config/vscode_config.rs
```diff
@@ -17,6 +15,8 @@ pub async fn get_client_config_vscode(
     context: &ServerContextSnapshot,
     config: &mut ClientConfig,
 ) -> Option<()> {
+    get_client_config_default(context, config, None).await;
```
CppCXY, line 18: Why request configuration? The existing configuration in EmmyLua is unrelated to the language server itself; it’s mainly for color settings.

taminomara: I want to provide a way to change non-project-related settings (like which inlay hints to display, whether to render markdown in comments or not, etc.) from VSCode interface. This approach repeats what rust-analyzer and several other LSPs do.

Current WIP on VSCode settings:

image

Plus, I want to provide a standard way for all editors to change EmmyLua's config.

crates/emmylua_ls/src/handlers/initialized/client_config/mod.rs
taminomara: I want to make it similar to VSCode. See PR description for detailed explanation:

> Currently, both LuaLs and EmmyLua use the same scope "Lua". It works in NeoVim because options for every LSP server are independent. In VSCode, however, they all share the same namespace, meaning that we can't reuse "Lua". I'd like to keep things consistent, hence this rename.

crates/emmylua_ls/src/handlers/initialized/client_config/mod.rs
taminomara: Added check for client capabilities.

crates/emmylua_ls/src/handlers/initialized/client_config/vscode_config.rs
CppCXY: Manual maintenance is very difficult. LuaLS generates its package.json, so if this is to be implemented, I hope it can be generated automatically.

taminomara: With schema export already present, this shouldn't be hard to implement. I'll add it when sending PR to the VSCode extension.

PR #692, opened 2025-08-07
fix(hover): use left token when cursor is before '.', ':', '[', ']'

skewb1k:
Fixes an issue with hover behavior on dot selectors. When the cursor was placed to the left of the dot (e.g., `struct|.field`), the hover would incorrectly show information for the field and highlight the dot token.

Before/After:
1754560858_screenshot
1754560904_screenshot

### DIFF
--- a/crates/emmylua_ls/src/handlers/hover/mod.rs
+++ b/crates/emmylua_ls/src/handlers/hover/mod.rs
@@ -9,7 +9,7 @@ mod std_hover;
 pub use build_hover::build_hover_content_for_completion;
 use build_hover::build_semantic_info_hover;
 use emmylua_code_analysis::{EmmyLuaAnalysis, FileId};
-use emmylua_parser::LuaAstNode;
+use emmylua_parser::{LuaAstNode, LuaTokenKind};
 pub use find_origin::{find_all_same_named_members, find_member_origin_owner};
 pub use hover_builder::HoverBuilder;
 pub use hover_humanize::infer_prefix_global_name;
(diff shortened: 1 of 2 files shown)


xuhuanzy:
You need to write a test for this to ensure correct execution.

arkriny:
Added test. Fails on main, passes with the fix.

crates/emmylua_ls/src/handlers/hover/mod.rs
```diff
@@ -56,10 +56,16 @@ pub fn hover(analysis: &EmmyLuaAnalysis, file_id: FileId, position: Position) ->
 
     let token = match root.syntax().token_at_offset(position_offset) {
         TokenAtOffset::Single(token) => token,
-        TokenAtOffset::Between(_, right) => right,
-        TokenAtOffset::None => {
-            return None;
+        TokenAtOffset::Between(left, right) => {
+            if LuaTokenKind::from(right.kind()) == LuaTokenKind::TkDot {
```
CppCXY: it should also include ':' and '[', ']', and you can use matches!:
```rust
if matches!(right.kind().into(), LuaTokenKind::TkDot ...)
```

crates/emmylua_ls/src/handlers/test/hover_test.rs
```diff
@@ -298,4 +298,23 @@ mod tests {
             },
         ));
     }
+
+    #[test]
+    fn test_before_dot_returns_object_info() {
+        let mut ws = ProviderVirtualWorkspace::new();
+        assert!(ws.check_hover(
```
CppCXY: emmylua_ls has introduced gtest, and you can use the check! macro here.

crates/emmylua_ls/src/handlers/test/hover_test.rs
```diff
@@ -298,4 +298,23 @@ mod tests {
             },
         ));
     }
+
+    #[test]
```
CppCXY: gtest

PR #698, opened 2025-08-08
Reuse doc completion, hover, and definition for `@see`

### LINKED ISSUE #671: Make syntax of `@see` compatible with LuaLs
LuaLs uses `#` to split reference and description:

```lua
--- @see Foo.bar # description of `see` content.
```

while EmmyLua uses it to split type and member:

```lua
--- @see Foo#bar
```

# Considerations

In order not to break existing code, we can check whether there are spaces around `#`. Alternatively, we can drop current behavior and only use LuaLs syntax.

taminomara:
This PR improves handling of `@see` contents by reusing machinery introduced for parsing documentation comments. The behavior is backwards-compatible with the old syntax (`@see class#member`).

Fix #671

### DIFF
--- a/crates/emmylua_code_analysis/src/compilation/analyzer/doc/type_ref_tags.rs
+++ b/crates/emmylua_code_analysis/src/compilation/analyzer/doc/type_ref_tags.rs
@@ -389,11 +389,16 @@ pub fn analyze_see(analyzer: &mut DocAnalyzer, tag: LuaDocTagSee) -> Option<()> 
     let owner = get_owner_id_or_report(analyzer, &tag)?;
     let content = tag.get_see_content()?;
     let text = content.get_text();
+    let descriptions = tag
+        .get_description()
+        .map(|description| description.get_description_text());
 
-    analyzer
-        .db
-        .get_property_index_mut()
-        .add_see(analyzer.file_id, owner, text.to_string());
+    analyzer.db.get_property_index_mut().add_see(
+        analyzer.file_id,
+        owner,
+        text.to_string(),
+        descriptions,
+    );
 
     Some(())
 }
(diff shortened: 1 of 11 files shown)


crates/emmylua_parser/src/lexer/lua_doc_lexer.rs
```diff
@@ -403,7 +403,30 @@ impl LuaDocLexer<'_> {
                 LuaTokenKind::TkWhitespace
             }
             _ => {
-                reader.eat_while(|_| true);
+                let mut brace_level = 0usize;
```
CppCXY: can not change the lexer, Because anything can follow after "see".

taminomara: On the contrary, this change allows handling anything after `@see`, not only type name. If user follows LuaLs syntax, this will be parsed correctly so that completion and go to definition works. If not, `@see` content will just be a description, which is not structured.

CppCXY: if I write `---@see Program Files/...`, So I can only see `Program`  in the hover, and after splitting by spaces, their semantic tokens are also different

taminomara: Alright, I see. On the other hand, you can enclose file names in quotes, and maybe in future we'll implement autocomplete for resources for such cases. Overall, I feel like adding more structure to `@see` content is a good thing. This will also help with documentation generators, because now they can't do anything meaningful with `@see`, only dump it as-is.

Or I can undo this change, it's your call.

CppCXY: see is a commonly used tag, but its format is highly variable. Structuring it during the parser phase doesn’t make much sense, and it would also break the export of existing see content.

PR #699, opened 2025-08-09
Add option to increase verbosity of types rendered in hovers

taminomara:
Fix EmmyLua/VSCode-EmmyLua#250


taminomara:
Ping?

CppCXY:
I think the modification direction is incorrect. You should change Documentation to Custom(usize), and then pass the corresponding value through configuration.

CppCXY:
- [ ]

PR #705, opened 2025-08-11
Remove module nesting in tests

taminomara:
Since the entire tests directory is now under `cfg(test)`, we don't need
an additional `test` module for every file.

This PR only changes indentation, no logic is changed. I've added `.git-blame-ignore-revs` so that git doesn't show it in `git blame`. You can enable it with this command:

```sh
git config blame.ignoreRevsFile .git-blame-ignore-revs
```


CppCXY:
There’s no need, because I checked and all the tests are written this way. I don’t want to modify all of them.

PR #706, opened 2025-08-11
Switch `emmylua_parser_desc` to gtest

taminomara:
GTest prints nice diffs, which is useful for debugging. It also allows continuing test even if assertion fails, so you see all failures if several test cases are crammed in a single `#[test]`.

image


CppCXY:
please resolve the conflicts

PR #708, opened 2025-08-12
Warn and ignore variadic expansions that can't be handled properly

### LINKED ISSUE #701: Improve semantics of variadic expansion
Right now, we allow variadic type expansion anywhere in a type expression. This leads to potential unsoundness where `Variadic<T>` are not properly unwrapped:

```lua
local t --- @type int...
local a, b, c, d = t
-- `a`, `b`, `c`, `d` are all `int`.

local t --- @type [int..., string]
local a, b, c, d = t[1]
-- `a`, `b`, `c`, `d` are all `int`.
local e = t[2]
-- `e` is `string`
```

I'd like to fix this.

There are two approaches we could take:

1. disallow variadic expansion in places where it can't be handled.

   Specifically, using `T...` will be allowed:
   - in last element of a tuple, i.e. `[int, string...]`,
   - in generic arguments, i.e. `Foo<T...>` (semantics of this needs revision, but I'll leave it for later),
   - on top level of function/operator parameters and return values, i.e. `@param ... T...`.
   
   The rest will be reported as an error.

2. handle variadics everywhere, with default behavior to extract the first value.

   This will make variadic expansion a no-op outside of the cases mentioned above.

   Specifically, `int...` will be treated as `int`, `T...` will be treated as `T`, and so on.

I like option 1. Option 2 is more error prone, and can lead to unexpected behavior. For example, consider this:

```lua
--- @generic T
--- @param ... T...
--- @return [T..., table]
function foo(...) end

local x = foo(1, "")
```

What should the type of `x` be? Is it `[integer, string, table]`, `[integer, table]`, or `[integer, string]`? Current implementation derives `[integer, string]`, but is it reasonable?

taminomara:
This PR checks for situation where variadic type expansion operator (`T...`) is used in an unexpected place, and can't be handled properly. Users will get a warning about such situations, and the operator will be ignored.

This prevents unexpected variadic types such as:

```lua
local x --- @type integer...
local x1, x2 = x
--             ^ EmmyLua thinks `x` is a result of a variadic expansion,
--    ^^  ^^     so it infers both `x1` and `x2` as integers.
```

Fix #701

### DIFF
--- a/crates/emmylua_code_analysis/src/compilation/test/variadic_test.rs
+++ b/crates/emmylua_code_analysis/src/compilation/test/variadic_test.rs
@@ -1,0 +1,224 @@
+#[cfg(test)]
+mod test {
+    use crate::{DiagnosticCode, VirtualWorkspace};
+
+    #[test]
+    fn test_unexpected_variadic_expansion() {
+        let mut ws = VirtualWorkspace::new();
+
+        assert!(!ws.check_code_for(
+            DiagnosticCode::DocTypeUnexpectedVariadic,
+            r#"
+                --- @type integer...
+                local _
+            "#,
+        ));
+
+        assert!(!ws.check_code_for(
+            DiagnosticCode::DocTypeUnexpectedVariadic,
+            r#"
+                --- @type (integer...)[]
+                local _
+            "#,
+        ));
+
+        assert!(!ws.check_code_for(
+            DiagnosticCode::DocTypeUnexpectedVariadic,
+            r#"
+                --- @type [integer..., integer]
+                local _
+            "#,
+        ));
+
+        assert!(!ws.check_code_for(
+            DiagnosticCode::DocTypeUnexpectedVariadic,
+            r#"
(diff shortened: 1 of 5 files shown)


crates/emmylua_code_analysis/src/compilation/test/variadic_test.rs
```diff
@@ -0,0 +1,224 @@
...
+            r#"
+                --- @class Foo
+                --- @operator call(Foo): Foo...
+            "#,
+        ));
+
+        assert!(ws.check_code_for(
+            DiagnosticCode::DocTypeUnexpectedVariadic,
+            r#"
+                --- @parameter x fun(integer...)
```
CppCXY: wrong annotation

taminomara:
Ping?

CppCXY:
code conflicts

taminomara:
Sorry, I was busy with other tasks, I'll take a look at it tomorrow.

PR #709, opened 2025-08-12
Prevent double variadic wrapping when inferring `...`

taminomara:
This prevents situation when `...` is annotated as variadic (`@param ... T...`), and its type is inferred as `(T...)...`

### DIFF
--- a/crates/emmylua_code_analysis/src/compilation/test/dots_infer_test.rs
+++ b/crates/emmylua_code_analysis/src/compilation/test/dots_infer_test.rs
@@ -1,0 +1,80 @@
+#[cfg(test)]
+mod test {
+    use crate::{FileId, LuaType, VirtualWorkspace};
+    use emmylua_parser::LuaLocalStat;
+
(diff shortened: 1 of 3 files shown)


crates/emmylua_code_analysis/src/compilation/test/dots_infer_test.rs
```diff
@@ -0,0 +1,80 @@
...
+            r#"
+                --- @generic T
+                --- @param ... T...
+                function foo(...)
+                    local a = { ... }
+                end
+            "#,
+        );
+        let a_ty = get_type_of_first_local_assign(&mut ws, file_id);
+        assert_eq!(&ws.humanize_type(a_ty), "T[]");
```
CppCXY, line 78: This is not the intended usage.

taminomara: I know, but I'm not sure how to compare types properly without hardcoding `GenericTplId`s?

taminomara:
Ping?

PR #710, opened 2025-08-12
Prevent `TplRef`s from escaping into types of global variables

taminomara:
If a global variable is not annotated, and there's an assignment to it in a generic function, there's a chance that references to generic parameters will end up in its type.

This commit fixes the situation by skipping a type from a particular assignment if it contains generic references.

This is a rather simplistic implementation. We could be smarter about it: it's possible to analyze generics that're referenced from a type and replace them with their bounds:

```lua
--- @generic T: Bound
--- @param x T
function foo(x)
   -- After this commit, `globalVar` will have type `undefined` in other files;
   -- instead, we can infer it as `Bound`.
   globalVar = x
end
```

However, this will require saving bounds to a global context, and lazily replacing them later. I've decided to leave it as is for now.

### DIFF
--- a/crates/emmylua_code_analysis/src/compilation/test/flow.rs
+++ b/crates/emmylua_code_analysis/src/compilation/test/flow.rs
@@ -1,6 +1,5 @@
 #[cfg(test)]
 mod test {
-
     use crate::{DiagnosticCode, LuaType, VirtualWorkspace};
 
     #[test]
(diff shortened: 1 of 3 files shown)


taminomara:
A particularly interesting scenario, where `TplRef`s from one function get evaluated in another:

```lua
--- @generic T
--- @param x T
function foo(x)
    -- Type of `globalVar` will be `TplRef(0)`.
    globalVar = x
end

--- @generic U
--- @param x U
function bar(x)
    -- EmmyLua picks up type of `globalVar`, which is `TplRef(0)`.
    -- Hence, function return type is also `TplRef(0)`.
    -- The signature of `bar` is therefore `bar<U>(x: U) -> U`.
   return globalVar
end

local result = bar(10)
--    ^^^^^^ type of `result` is `integer`
```

PR #755, opened 2025-09-10
Expose reachable types

Nerixyz:
A few types were not marked `pub` even though they could be reached through public types. For example, many returned references in methods from `DbIndex` were not exposed, so [they showed up gray](https://docs.rs/emmylua_code_analysis/0.13.0/emmylua_code_analysis/struct.DbIndex.html) in the docs. This also meant users of the crates couldn't spell the types.

I thought there was a clippy lint for this, but I couldn't find it. Also, I noticed that `cargo clippy` emits more than 1000 warnings and a few errors in this workspace. I could help with fixing them.


CppCXY:
We didn't actually use Clippy; we just made sure there were no warnings during compilation.

PR #818, opened 2025-10-25
fix(config): proper tilde expansion

### LINKED ISSUE #815: `~`->`$HOME` expansion not working
If I have this in my `.luarc.json` and/or `.emmyrc.json`:
```json
"workspace.library": [
	"~/Downloads/kristal-lua-docs",
	"~/Downloads/Kristal/luadoc_meta",
	"~/Downloads/luaCATS-love2d-def"
]
```
It doesn't recognize it. But if I have this instead:
```json
"workspace.library": [
	"/home/rudxain/Downloads/kristal-lua-docs",
	"/home/rudxain/Downloads/Kristal/luadoc_meta",
	"/home/rudxain/Downloads/luaCATS-love2d-def"
]
```
It works

Rudxain:
- [x] fixes  #815
- [ ] adds tests, to ensure it never happens again
- [ ] minor refactor (optional)

### DIFF
--- a/crates/emmylua_code_analysis/src/config/mod.rs
+++ b/crates/emmylua_code_analysis/src/config/mod.rs
@@ -149,7 +149,7 @@ fn pre_process_path(path: &str, workspace: &Path) -> String {
                 return path;
             }
         };
-        path = home_dir.join(&path[1..]).to_string_lossy().to_string();
+        path = home_dir.join(&path[2..]).to_string_lossy().to_string();
     } else if path.starts_with("./") {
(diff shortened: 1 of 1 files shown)


CppCXY:
What else is keeping it in draft?

Rudxain:
well, it can be merged as-is. But, to add tests, I guess I need to "modularize" `pre_process_path` in order to make it "pure", so that it can be tested in isolation from an OS. Right now, the fn depends on data that comes "from outside" (such as `home_dir`)

Rudxain:
I was going to ask if I should split the PR. I'll take that as a "yes" :sweat_smile:

Issue #765, opened 2025-09-15
`workspace/didChangeConfiguration` handling


lkubb:
First of all, thanks for your tireless work on this very helpful project!

Context: I'm using nvim + https://github.com/folke/lazydev.nvim (with a minor patch for emmylua support) to work on neovim plugins and my neovim configuration. When a buffer is loaded or changed, `lazydev` looks for `---@module "foo.bar"`/`require "foo.bar"` lines and dynamically adds the referenced modules to the workspace libraries. It does this by sending a `workspace/didChangeConfiguration` notification and handling the following `workspace/configuration` request.

Example code:
```lua
local lazydev = require("lazydev")
---@type lazydev.Config
local opts = {}
lazydev.setup(opts)
```
I'm having two issues with emmylua's handling of this request:

1. Since `v0.11.0` (likely https://github.com/EmmyLuaLs/emmylua-analyzer-rust/commit/9464063492bd577b8fdbf9f985bde64928278aee), when opening the first lua file (which causes emmylua to be launched), no `workspace/configuration` request is sent by emmylua in response to lazyvim's notification and all dynamically added libraries are unknown. This worked fine in `v0.10.0` and below.
When I manually send the notification a bit later, emmylua does send the request and is able to discover the modules. I think this is caused by the server still being in the process of initialization when lazydev sends its notification, which causes it to be discarded [here](https://github.com/EmmyLuaLs/emmylua-analyzer-rust/blob/cb88050f2b715e0376ba8acd86ce685fe7db56e8/crates/emmylua_ls/src/handlers/configuration/mod.rs#L15-L17). Would it be possible to queue/debounce its processing instead?

2. Even though emmylua is able to discover the dynamically added modules after receiving the new configuration (in the above example, I am able to hover/goto definition for `lazydev.Config`), annotations are not updated (in the above example, the `@type` declaration is annotated with `Type 'lazydev.Config' not found`). Screenshot: <img width="674" height="116" alt="Image" src="https://github.com/user-attachments/assets/5963a59b-bd2b-4b20-9594-1532a5f37533" />
Reloading the buffer (`:edit`) or restarting the LSP client (`:LspRestart`) causes this annotation to be removed. I checked whether I could force emmylua to send fresh annotations via the `textDocument/diagnostic` request, but it's not supported by emmylua.

CppCXY:
[...]

PR #856, opened 2025-11-17
fix: don't drop inbound `workspace/didChangeConfiguration` notifications

Davidyz:
In #765 there were discussions about the first file opened in neovim won't have the updated config (`workspace.library` for dynamically adding libraries) sent in the `workspace/configuration` response. 

I believe this was caused by [the `init_analysis` function being called before the workspace manager has been initialized](https://github.com/EmmyLuaLs/emmylua-analyzer-rust/blob/978b68335d52ecadbbda61c0747a428fb4d40cda/crates/emmylua_ls/src/handlers/initialized/mod.rs#L80). This means the [workspace manager in the `workspace/configuration` handler](https://github.com/EmmyLuaLs/emmylua-analyzer-rust/blob/978b68335d52ecadbbda61c0747a428fb4d40cda/crates/emmylua_ls/src/handlers/configuration/mod.rs#L17) is incorrectly marked as uninitialized for the first file.

This PR address the issue, and when used in neovim and [lazydev.nvim](https://github.com/folke/lazydev.nvim), `emmylua_ls` can now correctly load the dynamically added libraries.

### DIFF
--- a/crates/emmylua_ls/src/server/message_processor.rs
+++ b/crates/emmylua_ls/src/server/message_processor.rs
@@ -30,10 +30,7 @@ impl ServerMessageProcessor {
             Message::Response(_) => true,
             // Allow specific notifications
             Message::Notification(notify) => {
-                matches!(
-                    notify.method.as_str(),
-                    "workspace/didChangeConfiguration" | "$/cancelRequest" | "initialized"
-                )
+                matches!(notify.method.as_str(), "$/cancelRequest" | "initialized")
             }
             // Don't process other requests during initialization
             Message::Request(_) => false,


CppCXY:
This is not the correct way to fix the issue. This approach is essentially a race between different requests reaching the language server. It's entirely possible for workspace/configuration to reach the language server first, and then for workspace_manager to finish initialization afterwards.

CppCXY:
Actually, I know how to fix it, see here, remove the "workspace/didChangeConfiguration"
https://github.com/EmmyLuaLs/emmylua-analyzer-rust/blob/978b68335d52ecadbbda61c0747a428fb4d40cda/crates/emmylua_ls/src/server/message_processor.rs#L35

but I think initializing twice is really stupid.

CppCXY:
I think the place that should be fixed is here:
https://github.com/EmmyLuaLs/emmylua-analyzer-rust/blob/main/crates/emmylua_ls/src/handlers/initialized/client_config/default_config.rs

I have already proactively requested the configuration from the client, so why hasn't it taken effect?

Davidyz:
> This is not the correct way to fix the issue. This approach is essentially a race between different requests reaching the language server. It's entirely possible for workspace/configuration to reach the language server first, and then for workspace_manager to finish initialization afterwards.

Thanks for the comments, but I'm not totally convinced. In `init_analysis`, the `analysis` argument is protected by `RwLock` to prevent the race condition. The `file_diagnostic` argument also has internal locks/mutex that should handle the concurrency. I believe this is also why you didn't have to do anything special when calling `init_analysis` in [`reload_workspace`](https://github.com/EmmyLuaLs/emmylua-analyzer-rust/blob/978b68335d52ecadbbda61c0747a428fb4d40cda/crates/emmylua_ls/src/context/workspace_manager.rs#L142), which is triggered when there's an incoming `workspace/didChangeConfiguration` noti.

> but I think initializing twice is really stupid.

I agree that we should avoid initialize (that is, the `initialize` in the LSP lifecycle) twice.

> I have already proactively requested the configuration from the client, so why hasn't it taken effect?

This might have been a timing issue. In lazydev.nvim (a neovim plugin that dynamically updates `workspace.library`), the LSP handlers don't get registered from the beginning, but rather [when there are some changes in `workspace.library`](https://github.com/folke/lazydev.nvim/blob/5231c62aa83c2f8dc8e7ba957aa77098cda1257d/lua/lazydev/buf.lua#L204). This won't be done until some parsing has been done with the file and some libraries have been identified. This means, when emmylua_ls requests client config in `initialized_handler`, the `workspace/configuration` handler in neovim might not have been updated to correctly respond with the updated `workspace.library`.

CppCXY:
> Thanks for the comments, but I'm not totally convinced. In `init_analysis`, the `analysis` argument is protected by `RwLock` to prevent the race condition. The `file_diagnostic` argument also has internal locks/mutex that should handle the concurrency. I believe this is also why you didn't have to do anything special when calling `init_analysis` in [`reload_workspace`](https://github.com/EmmyLuaLs/emmylua-analyzer-rust/blob/978b68335d52ecadbbda61c0747a428fb4d40cda/crates/emmylua_ls/src/context/workspace_manager.rs#L142), which is triggered when there's an incoming `workspace/didChangeConfiguration` noti.

This might be a translation issue. When I mentioned "race," I didn't mean a race condition, but rather that the handling order of these two requests could be random. So it's still possible that when processing workspace/configuration, the workspace_manager's initialized flag hasn't been set yet. The reason why swapping the order currently works is because init_analysis takes quite some time to process. I can accept initializing twice on lazyvim, but I can't accept this change affecting other platforms, because in the game environment there could be tens of thousands of Lua files, and the cost of initializing twice is too high.

Davidyz:
> The reason why swapping the order currently works is because init_analysis takes quite some time to process. I can accept initializing twice on lazyvim, but I can't accept this change affecting other platforms, because in the game environment there could be tens of thousands of Lua files, and the cost of initializing twice is too high.

I take it that when you say "initializing twice", you're referring to the `init_analysis` function being called twice. There are 2 ways that the `init_analysis` function can be called:

1. in the `initialized_handler`. This will, by definition, only be called once at the start of the LSP lifecycle.
2. In the handler of a client-to-server notification that calls `workspace_manager.reload_workspace` or `workspace_manager.add_update_emmyrc_task`.

Since `1` can only occur once, we can conclude that _if `init_analysis` is called more than once in a very short time frame, it's because the client decided to do so_. Regardless of whether [the `workspace/didChangeConfiguration` is quietly thrown away](https://github.com/EmmyLuaLs/emmylua-analyzer-rust/blob/978b68335d52ecadbbda61c0747a428fb4d40cda/crates/emmylua_ls/src/handlers/configuration/mod.rs#L23), the client can trigger frequent `init_analysis` calls by sending notifications that trigger the relevant handlers, and the existing code here won't help with that anyways. I also don't think it's a good idea to just quietly throw away notifications for performance. This eventually makes the LSP operate in an outdated (and therefore incorrect) state.

CppCXY:
This is the correct method but not the appropriate one. This issue is introduced by lazyvim and is not a problem with all editors. If every editor correctly sent workspace/configuration, I could simply wait for its result before starting initialization. However, that's not the case—some editors may not support workspace/configuration. I can't rely on this behavior for my initialization.

Now there is another method: adding a dedicated `--editor lazyvim.dev`. In this case, workspace/configuration is added to the pending requests, which results in two initializations.

Davidyz:
> This issue is introduced by lazyvim and is not a problem with all editors. If every editor correctly sent workspace/configuration

I think that's unfair. This issue is caused by the workspace manager not fully initialized after the server sends the `InitializeResult`. This means the server isn't actually "initialized" when it sends the `InitializeResult`. [The LSP specification only mentions that the client should not send other requests/notifications before it receives the `InitializeResult`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initialize), which neovim/lazydev did not violate. There's no way for the clients to tell when the workspace manager is actually initialized and is ready to process the `didChangeConfiguration` notifications.

> some editors may not support workspace/configuration

May I propose another approach: In `initialized_handler`, we use `param.capabilities` to determine whether the client supports the `workspace/configuration` request (maybe also the `didChangeConfiguration` notification). If the client doesn't support it, we use the original order (`init_analysis` first, initialize `workspace_manager` afterwards) because we don't have to worry about the config changes. If the client supports it, we use the order in this PR so that emmylua correctly listens to the configuration updates.

Update: tbh this alternative approach is still not perfect because there can still be incoming `didChangeConfiguration` notifications between `initialize_finish` and workspace manager initialization. The most "correct" approach would be to either queue the incoming notification in the `on_did_change_configuration` handler until the workspace manager initialization finish, or to move the workspace manager initialization to before `connection.initialize_finish`. Either way, the notification sent after the LSP `InitializeResult` is sent shouldn't be ignored.

CppCXY:
I realized it's actually not that complicated, because in practice, other editors don't actively send workspace/configuration requests. The current check for initialized in workspace/configuration is an outdated solution, since I have modified the main_loop implementation to allow pending requests. Therefore, the initialized field in workspace_manager can be removed and allow pending the `workspace/configuration` logic .

Davidyz:
@CppCXY I've removed the `workspace_initialized` flag and the relevant getter/setter. The `init_analysis` call in `initialize_handler` still needs to be after `workspace_manager.client_config` is updated to make the first analysis run with the up-to-date information.

CppCXY:
I think you can remove `workspace/configuration` from [emmylua-analyzer-rust/crates/emmylua_ls/src/server/message_processor.rs](https://github.com/EmmyLuaLs/emmylua-analyzer-rust/blob/978b68335d52ecadbbda61c0747a428fb4d40cda/crates/emmylua_ls/src/server/message_processor.rs#L35)

Davidyz:
There might still be something wrong with this PR. Now I sometimes don't get the updated workspace library at all (even after opening the second file). From the logs it seems like the server was able to pull the client configs.

Davidyz:
@CppCXY I'm noticing deadlocks in `init_analysis`. Specifically, `analysis.write().await` hangs when it's triggered by reload workspace. The `initialized_handler` can finish just fine. The deadlock doesn't always occur, so I missed it before the merge.

Davidyz:
I was just poking around the notification handlers that are triggered at the beginning of a session, and flipping the order of [`analysis().read()` and `workspace_manager().read()`](https://github.com/EmmyLuaLs/emmylua-analyzer-rust/blob/4d6f7dae7be3509fa6c0dc58d25317222f4e36df/crates/emmylua_ls/src/handlers/text_document/text_document_handler.rs#L19) seems to fix the deadlock (at least the following `init_analysis` calls are no longer blocked).

CppCXY:
can you make a pr fix this?

Davidyz:
I'm still trying to figure out the underlying mechanism of the deadlock. I encountered some (very rare) deadlocks even with the previously mentioned "fix", so it's probably not the actual culprit. I don't want to submit another non-solution.

btw if you want to revert this PR due to the deadlock, I'm ok with it.

CppCXY:
> I'm still trying to figure out the underlying mechanism of the deadlock. I encountered some (very rare) deadlocks even with the previously mentioned "fix", so it's probably not the actual culprit. I don't want to submit another non-solution.
> 
> btw if you want to revert this PR due to the deadlock, I'm ok with it.

I think we can simply restore the use of the initialized field, because it was originally designed to solve deadlocks.

Davidyz:
> I think we can simply restore the use of the initialized field, because it was originally designed to solve deadlocks.

I think that'll bring back the "discarding notification" issue because `on_did_change_configuration` is the only place that you read the initialized flag for doing things. It'll be just like reverting this PR.

CppCXY:
I try fix this deadlock: b3d565e8ecc5605388d0f890c5a164b19c1eb9e7
please check is this work

Davidyz:
@CppCXY seems to work fine. Thanks.

btw, the following is what I've come up with about the deadlock issue: 

1. when `init_analysis` is called, it [acquires a write lock on `analysis`](https://github.com/EmmyLuaLs/emmylua-analyzer-rust/blob/b3d565e8ecc5605388d0f890c5a164b19c1eb9e7/crates/emmylua_ls/src/handlers/initialized/mod.rs#L111), holds the lock and [create a progress bar](https://github.com/EmmyLuaLs/emmylua-analyzer-rust/blob/b3d565e8ecc5605388d0f890c5a164b19c1eb9e7/crates/emmylua_ls/src/handlers/initialized/mod.rs#L121). During the progress bar creation, it [sends a request to the client and await for the response](https://github.com/EmmyLuaLs/emmylua-analyzer-rust/blob/b3d565e8ecc5605388d0f890c5a164b19c1eb9e7/crates/emmylua_ls/src/context/status_bar.rs#L48).
2. Before the response arrives, a request arrives, and the corresponding handler tries to acquire [a lock on `analysis`](https://github.com/EmmyLuaLs/emmylua-analyzer-rust/blob/b3d565e8ecc5605388d0f890c5a164b19c1eb9e7/crates/emmylua_ls/src/handlers/inlay_hint/mod.rs#L21) while blocking the message loop in the `run` function in `lsp_server`.

Since async notification handlers are spawned to run concurrently, they can run in parallel with the message loop in `lsp_server`, and the 2 paths mentioned above might result in a deadlock.

I think finer control over the progress bar might help solve this issue. The current implementation also doesn't care whether the client supports progress bar capabilities. If it doesn't (or if the user decided to disable it), and therefore doesn't send a reply to the `window/workDoneProgress/create` request, the `create_progress_task` function may hang forever.

PR #861, opened 2025-11-23
feat(emmylua_ls): show progress percentage while indexing workspace

Davidyz:
This PR adds support for sending percentage progress when initializing the codebase.

### DIFF
--- a/crates/emmylua_code_analysis/src/lib.rs
+++ b/crates/emmylua_code_analysis/src/lib.rs
@@ -156,6 +156,37 @@ impl EmmyLuaAnalysis {
         updated_files
     }
 
+    pub fn update_files_by_uri_with_progress<T: Fn(usize, usize)>(
+        &mut self,
+        files: Vec<(Uri, Option<String>)>,
+        progress_callback: T,
+    ) -> Vec<FileId> {
+        let mut removed_files = HashSet::new();
+        let mut updated_files = HashSet::new();
+        let total_count = files.len();
+        {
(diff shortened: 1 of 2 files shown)


CppCXY:
This is an incorrect use of the API. Loading all files together and performing out-of-order analysis aligns with the reality of Lua. Introducing progress reporting and loading them separately not only reduces performance but also leads to incorrect results.

Davidyz:
I see. I have another design: create a new function `EmmyLuaAnalysis::update_files_by_uri_with_progress` that accepts a callback. We can then pass a closure function that updates the progress bar. The `update_files_by_uri_with_progress` function can use the same design as the other batch update functions and handle the `update_index` and `remove_index` altogether. Would that be a acceptable solution?

CppCXY:
Your current progress report only monitors the construction of the syntax tree, which constitutes a very small part of the analysis process. The time spent reporting progress may far exceed the time required to build the syntax tree. This is an unacceptable performance degradation. Progress bars should not be used to monitor performance-intensive analysis stages.

Davidyz:
Indeed. I noticed that the progress will stay at 100% for a while, which means the progress representation is inaccurate.

> Progress bars should not be used to monitor performance-intensive analysis stages.

The progress percentage is a notification, so it doesn't need to wait for a response, and therefore shouldn't be a performance bottleneck. Also, if a task doesn't take very long, there's no point of using a progress bar, because you'll hardly be able to see the "progress".

PR #919, opened 2026-01-06
enhance os library annotations

tinyleolin:


clason:
So there's a bit of an issue with this change: if I do `debug.getinfo(1, 'S')`,
* `emmylua_check` will warn that this could be `nil`, but
* `emmylua` will warn that an assert or conditional will always be truthy.

So I'll get warnings either way. How do you suggest this should be handled?

CppCXY:
> So there's a bit of an issue with this change: if I do `debug.getinfo(1, 'S')`,
> 
> * `emmylua_check` will warn that this could be `nil`, but
> * `emmylua` will warn that an assert or conditional will always be truthy.
> 
> So I'll get warnings either way. How do you suggest this should be handled?


Because the version number was not updated, the resources actually haven't been updated. You might need to delete the resource directory and let it regenerate.

clason:
I don't understand; I build and run both tools straight from the project directory. That has always worked fine.

(But for the sake of testing, I wiped all `.local/` directories; no difference.)

The issue here is that the language server and the CLI tool check the code in different contexts.

CppCXY:
> I don't understand; I build and run both tools straight from the project directory. That has always worked fine.
> 
> (But for the sake of testing, I wiped all directories; no difference.)`.local/`
> 
> The issue here is that the language server and the CLI tool check the code in different contexts.

emmylua-ls generates the API files to $XDG/emmylua_ls/resources during startup. However, this generation does not always occur. If files exist in this directory and there is a version file inside, the files will only be regenerated if that version differs from the version written into Cargo during the compilation of emmylua-ls.

clason:
Ok, thank you for the explanation.

PR #938, opened 2026-01-30
WIP: update

lewis6991:


xuhuanzy:
This PR seems to make heavy use of AI? Are you sure you fully understand what the changes mean?

lewis6991:
Not fully, but I've made sure every single change is backed up with a test, which constitutes most of the PR.

xuhuanzy:
Some edge cases are hard to cover with tests, so we can’t guarantee that passing all tests means it’s correct. Before making changes, it’s best to make sure you know what you’re doing.

lewis6991:
I'll run it through my projects before this is ready.

Otherwise I can break this down into easy to review PR's. The changes began simple but kept finding existing issues.

lewis6991:
changes applied to other PR's

PR #940, opened 2026-01-31
feat(diagnostic): add non-callable call checker

lewis6991:


crates/emmylua_code_analysis/src/diagnostic/test/call_non_callable_test.rs
```diff
@@ -0,0 +1,135 @@
...
+        assert!(ws.check_code_for(
+            DiagnosticCode::CallNonCallable,
+            r#"
+                ---@type function
+                local f
+                f()
+            "#
+        ));
+
+        assert!(!ws.check_code_for(
```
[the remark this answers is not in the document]
lewis6991: Keeping this as a diagnostic is intentional: we warn if the type includes any non-callable member (excluding nil), so function|integer should still trigger CallNonCallable.

crates/emmylua_code_analysis/src/diagnostic/test/call_non_callable_test.rs
```diff
@@ -0,0 +1,135 @@
...
+        ));
+
+        assert!(ws.check_code_for(
+            DiagnosticCode::CallNonCallable,
+            r#"
+                (function() end)()
+            "#
+        ));
+
+        assert!(!ws.check_code_for(
```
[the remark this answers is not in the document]
lewis6991: Same reasoning as above: the diagnostic is expected when a union contains a non-callable member, even if it also contains a callable one.

crates/emmylua_code_analysis/src/diagnostic/checker/call_non_callable.rs
```diff
@@ -0,0 +1,195 @@
...
+
+    match typ {
+        LuaType::Any | LuaType::Unknown | LuaType::SelfInfer | LuaType::Global | LuaType::Nil => {
+            false
+        }
+        LuaType::Union(union) => union.into_vec().iter().any(is_definitely_non_callable),
+        LuaType::MultiLineUnion(union) => union
+            .get_unions()
+            .iter()
+            .any(|(t, _)| is_definitely_non_callable(t)),
```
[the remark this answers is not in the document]
lewis6991: The intent here is to flag calls when the type includes any non-callable member (except nil), even if some members are callable. So any() is deliberate, and I’ve renamed the helper to reflect that semantic.

crates/emmylua_code_analysis/src/diagnostic/checker/call_non_callable.rs
```diff
@@ -0,0 +1,195 @@
...
+        Ok(_) if !non_callable => return Some(()),
+        Ok(_) => {}
+        Err(reason) if reason.is_need_resolve() => return Some(()),
+        Err(reason)
+            if !matches!(reason, InferFailReason::None | InferFailReason::RecursiveInfer) =>
+        {
+            return Some(());
+        }
+        Err(_) => {}
+    }
```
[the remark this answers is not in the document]
lewis6991: I changed this to return early on Ok(_), since a successful inference means the call is valid.

crates/emmylua_code_analysis/src/diagnostic/checker/call_non_callable.rs
```diff
@@ -0,0 +1,211 @@
...
+
+use super::{Checker, DiagnosticContext};
+
+pub struct CallNonCallableChecker;
+
+impl Checker for CallNonCallableChecker {
+    const CODES: &[DiagnosticCode] = &[DiagnosticCode::CallNonCallable];
+
+    fn check(context: &mut DiagnosticContext, semantic_model: &SemanticModel) {
+        for node in semantic_model.get_root().descendants::<LuaAst>() {
```
CppCXY: should directly use `LuaCallExpr` instead of `LuaAst`:
```rust
 semantic_model.get_root().descendants::<LuaCallExpr>()

```

crates/emmylua_code_analysis/src/diagnostic/checker/call_non_callable.rs
```diff
@@ -0,0 +1,209 @@
...
+            for t in union.into_vec() {
+                insert_if_non_callable(&t);
+            }
+        }
+        LuaType::MultiLineUnion(union) => {
+            for (t, _) in union.get_unions().iter() {
+                insert_if_non_callable(t);
+            }
+        }
+        LuaType::TypeGuard(inner) => {
```
CppCXY: typeguad is actually treated as a boolean

crates/emmylua_code_analysis/src/diagnostic/checker/call_non_callable.rs
```diff
@@ -0,0 +1,209 @@
...
+    let mut cache = semantic_model.get_cache().borrow_mut();
+    let call_result = infer_call_expr_func(
+        db,
+        &mut cache,
+        call_expr.clone(),
+        call_expr_type.clone(),
+        &InferGuard::new(),
+        None,
+    );
+    if let Err(reason) = &call_result {
```
CppCXY, line 42: Should use ".ok()?" directly?

lewis6991: How? I need access to `reason`.

CppCXY: Okay, it just feels like the logic here is redundant — why are there two if statements?

PR #946, opened 2026-02-03
fix(analysis): resolve alias-call returns and simplify flow assignments

lewis6991:
- infer_call: infer and apply alias-call substitutions from call-site
  arguments, including colon/dot alignment, so returns like std.Unpack
  are concrete at the call site.
- narrow/get_type_at_flow: drop explicit-annotation precedence and
  antecedent narrowing for assignments; just use the resolved expression
  type.
- stats: prefer declared global type caches for name prefixes so member
  owners attach to their type defs (e.g., stdlib globals).
- tests: add unpack alias-call coverage and a flow regression for
  doc-function assignment narrowing.

### DIFF
--- a/crates/emmylua_code_analysis/src/compilation/test/flow.rs
+++ b/crates/emmylua_code_analysis/src/compilation/test/flow.rs
@@ -259,6 +259,71 @@ print(a.field)
     }
 
     #[test]
+    fn test_doc_function_assignment_narrowing0() {
+        let mut ws = VirtualWorkspace::new();
+
+        let code = r#"
+        local i --- @type integer|fun():string
+        i = "str"
+        A = i
+        "#;
+
+        ws.def(code);
+        let a = ws.expr_ty("A");
+        let a_desc = ws.humanize_type_detailed(a);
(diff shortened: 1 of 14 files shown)


crates/emmylua_code_analysis/src/compilation/analyzer/lua/stats.rs
```diff
@@ -207,6 +207,44 @@ fn set_index_expr_owner(analyzer: &mut LuaAnalyzer, var_expr: LuaVarExpr) -> Opt
 
     match analyzer.infer_expr(&prefix_expr.clone()) {
         Ok(prefix_type) => {
+            // Prefer declared global types for name prefixes when choosing a member owner.
+            // This keeps stdlib members (like table.unpack) attached to their type defs.
+            let prefix_type = if let LuaExpr::NameExpr(name_expr) = &prefix_expr {
```
lewis6991: TODO: need to add test coverage for this block.

lewis6991: moved to #953

PR #947, opened 2026-02-04
fix(typecheck): accept intersections in class assignments

lewis6991:
- allow intersection types to be checked against class refs
- add tests for intersection<->class assignment behavior
- update intersection docs (EN/CN) and overload docs
- keep annotation tests aligned with intersection semantics

### DIFF
--- a/crates/emmylua_code_analysis/src/compilation/test/annotation_test.rs
+++ b/crates/emmylua_code_analysis/src/compilation/test/annotation_test.rs
@@ -191,6 +191,92 @@ mod test {
     }
 
     #[test]
+    fn test_merge_right_mapped_type() {
+        let mut ws = VirtualWorkspace::new();
+
+        ws.def(
+            r#"
+        ---@generic T: table, U: table
+        ---@param a T
+        ---@param b U
+        ---@return Merge<T, U>
+        local function extend(a, b) end
+
+        ---@type { x: number, y: number }
+        local a = { x = 1, y = 2 }
+
+        ---@type { y: string, z: boolean }
+        local b = { y = "hello", z = true }
+
+        c = extend(a, b)
+        "#,
+        );
+
+        let ty = ws.expr_ty("c.y");
+        let expected = ws.ty("string");
+        assert_eq!(ty, expected);
+    }
+
+    #[test]
+    fn test_extend_with_policy_overwrite() {
+        let mut ws = VirtualWorkspace::new();
+
+        ws.def(
+            r#"
+        ---@overload fun<T: table, U: table>(a: T, b: U, policy: "never"): T & U
+        ---@overload fun<T: table, U: table>(a: T, b: U, policy: "right"): Merge<T, U>
+        ---@overload fun<T: table, U: table>(a: T, b: U, policy: "left"): Merge<U, T>
+        ---@generic T: table, U: table
+        ---@param a T
+        ---@param b U
+        ---@param policy "never" | "left" | "right"
+        ---@return (T & U) | Merge<T, U> | Merge<U, T>
+        local function extend(a, b, policy) end
+
+        ---@type { x: number, y: number }
+        local a = { x = 1, y = 2 }
+
+        ---@type { y: string, z: boolean }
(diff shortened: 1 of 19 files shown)


xuhuanzy:
You should install the Git hooks first. The hooks include some basic checks to prevent failures in basic GitHub Actions checks.

Also, regarding AI usage, I recommend using `gpt-5.2` in Codex (don’t use `gpt-5.2-codex`), with the reasoning effort set to `xhigh`. For backend work, Codex’s capabilities far exceed CC.

lewis6991:
Ah ok, I have been using `gpt-5.2-codex`. I'll try `gpt-5.2`.

xuhuanzy:
`gpt-5.2-codex` trades quality for speed, and its output quality is very poor. It should only be used for asking questions, not for generating code.

xuhuanzy:
You can use `gpt-5.2` to improve the previous PR.

From what I’ve seen, code generated by `gpt-5.2` is usually two tiers better than `gpt-5.2-codex`, and most importantly, it tends to take edge cases into account.

lewis6991:
The previous PR already handles some pretty whacky edge cases (with tests). I can try though.

crates/emmylua_code_analysis/src/compilation/test/annotation_test.rs
```diff
@@ -190,6 +190,68 @@ mod test {
...
+
+        ---@type { y: string, z: boolean }
+        local b = { y = "hello", z = true }
+
+        c = extend(a, b)
+        "#,
+        );
+
+        let ty = ws.expr_ty("c.y");
+        let expected = ws.ty("string");
```
CppCXY, line 216: from ts it should be `never`:

Image

lewis6991: The current semantic that were implemented was LHS wins so `{a:integer} & {a:boolean}` would yield `{a:integer}`.

I thought this was a nice feature as it allows you to define a function like `merge` that usually has a merging policy. In neovim we have `vim.tbl_deep_extend('force'|'keep'|'error', tbl1, tbl2)` which is used a lot. `force` would yield `T2 & T1`, whilst `keep` would yield `T1 & T2`.

We could change this to `never`, but then how could you properly define a `merge` function?

lewis6991: Looks like in typescript, you need to implement this with:

```ts
type MergeRight<L, R> = Omit<L, keyof R> & R;

type Omit<T, K extends keyof any> = Pick<T, Exclude<keyof T, K>>;

type Pick<T, K extends keyof T> = { [P in K]: T[P]; };

type Exclude<T, U> = T extends U ? never : T;
```

or

```ts
type MergeRight<L, R> = { [P in Exclude<keyof L, keyof R>]: L[P] } & R;
````

CppCXY: We should indeed do that, but I still have some concerns about implementing complex generics — most of it is achieved through repeated recursion, so perhaps we should find a better approach.

lewis6991: Do you have any suggestions for this PR? Scale it back? Leave as is? Implement any of a above types directly without lower level generics?

For now I'd like if we could some way to accommodate the creation of `MergeRight/Left`, as that generally very is useful in Lua, or at least is in Neovim.

CppCXY: For now we can make this type built-in and support it with simple Rust logic(see: infer_special_generic_type), and change it later once we have sufficient infrastructure.

lewis6991: so implemented `MergeRight` and `MergeLeft`? And also remove the existing LHS wins semantics?

CppCXY: yes, this is the best approach for now.

PR #953, opened 2026-02-06
fix(typecheck): prefer declared globals for index member ownership

lewis6991:
When setting a member owner for an index expression with a name prefix
(e.g. `table.unpack`), prefer the declared global type cache over the
inferred prefix type. This keeps stdlib members attached to their type
defs and avoids losing them after casts (like `---@cast table unknown`).

Skip the override when the name is shadowed by a local.


lewis6991:
Not sure if this is something we care about, buts it's something codex picked up.

PR #962, opened 2026-02-23
fix: select(n, func()) correctly narrows type when func returns multiple values

nickvonkaenel:
## Summary

- `select(2, func())` where `func()` returns `(integer, string)` was resolving to `integer|string` instead of `string`
- Root cause: when a multi-return function call was the sole variadic argument, `variadic_tpl_pattern_match` bound the template `T` to the entire `Variadic(Multi([integer, string]))` wrapper rather than expanding it
- This caused `instantiate_variadic_type` to produce a doubly-nested `Variadic(Base(Variadic(Multi(...))))`, which bypassed the index-slicing logic in `instantiate_select_call`

## Fix

In `variadic_tpl_pattern_match` (`tpl_pattern/mod.rs`), when matching a `TplRef` variadic template against a single argument that is itself a `Variadic`, expand it instead of storing it as-is:

- `Variadic(Multi([T1, T2, ...]))` → `insert_multi_types([T1, T2, ...])` so `T...` instantiates to `Variadic(Multi([T1, T2, ...]))`
- `Variadic(Base(T))` → `insert_multi_base(T)` preserving unbounded variadic semantics

## Test plan

- [x] Added test case `select(n, multi_ret())` where `multi_ret()` returns `(integer, string)`, asserting result is `string`
- [x] All existing `test_select_type` cases still pass (variadic `...`, literal args, `select('#', ...)`)
- [x] Full test suite passes (570+ tests)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/emmylua_code_analysis/src/semantic/generic/tpl_pattern/mod.rs
+++ b/crates/emmylua_code_analysis/src/semantic/generic/tpl_pattern/mod.rs
@@ -803,9 +803,7 @@ pub fn variadic_tpl_pattern_match(
                             LuaType::Variadic(variadic) => match variadic.deref() {
                                 VariadicType::Multi(types) => match types.len() {
                                     0 => {
-                                        context
-                                            .substitutor
-                                            .insert_type(tpl_id, LuaType::Nil, true);
+                                        context.substitutor.insert_type(tpl_id, LuaType::Nil, true);
                                     }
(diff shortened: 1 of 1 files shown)


CppCXY:
please fix the check

nickvonkaenel:
Should be fixed now

PR #968, opened 2026-02-26
feat(diagnostic): report unresolved require modules

lewis6991:
Add an unresolved-require diagnostic for require("...") calls that
resolve to a string module path but cannot be found in the module index.

Extend the existing require visibility checker to emit
unresolved-require when module lookup fails.

### DIFF
--- a/crates/emmylua_code_analysis/resources/schema.json
+++ b/crates/emmylua_code_analysis/resources/schema.json
@@ -389,6 +389,11 @@
           "const": "cast-type-mismatch"
         },
         {
+          "description": "unresolved-require",
+          "type": "string",
+          "const": "unresolved-require"
+        },
+        {
           "description": "require-module-not-visible",
           "type": "string",
           "const": "require-module-not-visible"
--- a/crates/emmylua_code_analysis/src/diagnostic/checker/require_module_visibility.rs
+++ b/crates/emmylua_code_analysis/src/diagnostic/checker/require_module_visibility.rs
@@ -7,7 +7,10 @@ use super::{Checker, DiagnosticContext};
 pub struct RequireModuleVisibilityChecker;
 
 impl Checker for RequireModuleVisibilityChecker {
-    const CODES: &[DiagnosticCode] = &[DiagnosticCode::RequireModuleNotVisible];
+    const CODES: &[DiagnosticCode] = &[
+        DiagnosticCode::RequireModuleNotVisible,
+        DiagnosticCode::UnresolvedRequire,
+    ];
 
     fn check(context: &mut DiagnosticContext, semantic_model: &SemanticModel) {
         let root = semantic_model.get_root().clone();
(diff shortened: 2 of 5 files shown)


xuhuanzy:
This check is unnecessary; we can import `.dll` files or packages located outside the project path.  
If it must be included, it should be disabled by default.

lewis6991:
> This check is unnecessary; we can import .dll files or packages located outside the project path.

But then these are not type checked and you get `unknown` everywhere? If you are loading object files, then surely you must also have accompanying meta files to go with them?

xuhuanzy:
This is only an ideal scenario; in the current ecosystem, there exists a vast number of third-party libraries with undeclared types.

lewis6991:
What benefit do you get from this tool if libraries can't be resolved? So many things will not work: hover, goto, most diagnostics, etc. Having a single warning for each unresolved module shouldn't be that bad. It's certainly not unnecessary. And if it is that bad, what is so bad about downgrading the warning per project? I find it's much better for checks to be strict by default, rather than requiring users to enable everything. Selectively downgrading is much safer.

I understand that some environments do not have very good typing for the libraries they import, and that is why I made this a warning instead of an error.

xuhuanzy:
If the library path is incorrect, `lua` itself will report an error during execution, meaning we assume each module actually exists.

lewis6991:
Sorry I don't understand this point.

One of the main uses of emmylua (and all static analysis tools) is to catch issues statically (early) rather than at runtime (late).

xuhuanzy:
Have you considered the scenario when importing a `dll`? What about the case where `require` is overridden?
We cannot assume that this module does not exist. If it were absent, the runtime would definitely throw an error, which in turn implies that the module must exist.

xuhuanzy:
If you really want to join, then its warning level needs to be `Hint`.

CppCXY:
I think that since we're going to diagnose require, ambiguous require calls should also be detected. 
Requiring every C dynamic library to have a meta file is difficult, but we can, following the global approach, define a diagnostic exclusion list. With these two configuration options in place, I believe this diagnostic can be a warning — or even an error.

xuhuanzy:
> I think that since we're going to diagnose require, ambiguous require calls should also be detected. Requiring every C dynamic library to have a meta file is difficult, but we can, following the global approach, define a diagnostic exclusion list. With these two configuration options in place, I believe this diagnostic can be a warning — or even an error.

这样做那么你每个小脚本在引用到 luarocks 时都要建立一个 `.emmyrc.json`, 这很明显是不可能的, 最多 hint 得了, 有强需求才需要在 `.emmyrc.json` 提升到 error

lewis6991:
> Have you considered the scenario when importing a dll?

Meta files should be provided so the module can be resolved, even if the meta file just returns `any`. Otherwise, you can use `@diagnostic disable`. My codebases use `@diagnostic disable` for lots of cases, especially around dynamic code.

> What about the case where require is overridden?

If it's overridden, then surely it must still provide the same interface? Otherwise, the code is too dynamic and isn't very good to use with emmylua.

> We cannot assume that this module does not exist. If it were absent, the runtime would definitely throw an error, which in turn implies that the module must exist.

The check doesn't say the module doesn't exist, just that the checker cannot resolve it, which is important and worth having a warning for. It either indicates that the checkers environment isn't setup correctly, or there is a genuine typo in the module name.

lewis6991:
@CppCXY what can we do to progress this? Are there any changed you would like me to make?

CppCXY:
Considering that LuaRocks is rarely used in real projects (for example, I never use it),  emmylua supports a global .emmyrc.json configuration, so there is effectively no obstacle here.

PR #972, opened 2026-02-28
fix: check t[k] should not warn unnecessary-if

EfveZombie:


xuhuanzy:
This change needs a trade-off. I don’t think we need to handle `nil` here; adding it would introduce a lot of unnecessary annotations.

crates/emmylua_code_analysis/src/diagnostic/test/unnecessary_if_test.rs
```diff
@@ -26,4 +26,21 @@ mod test {
...
+    fn test_table_generic_index_not_unnecessary() {
+        let mut ws = VirtualWorkspace::new();
+        assert!(ws.check_code_for(
+            DiagnosticCode::UnnecessaryIf,
+            r#"
+        ---@type table<string, number>
+        local t = {}
+        local k = "hello"
+        local v = t[k]
+        if v then
```
lewis6991, line 40: I'd argue this is unnecessary. If you want it not to be then the type of `t` should be `table<string,number?>`

EfveZombie:
Current behaviors are same to `TypeScript` default mode (without `noUncheckedIndexedAccess`), so I was convinced

Thanks for reviewing /Or2

crates/emmylua_code_analysis/src/diagnostic/test/unnecessary_if_test.rs
EfveZombie: Thats what I did, quite counterintuitive to me, but u r right

PR #973, opened 2026-02-28
fix: std functions should match official manuals

EfveZombie:
## Summary of Changes

This PR systematically audits and corrects the EmmyLua type-annotation stubs under `crates/emmylua_code_analysis/resources/std/` so that parameter types, return types, version guards, and doc-comments accurately reflect the official Lua manuals (5.1 through 5.4).

---

### 1. `coroutine.lua` (+12 / −2)

| Item | Change | Rationale |
|---|---|---|
| `coroutine.isyieldable(co)` | Added the `co` parameter name to the function signature | Lua 5.4 manual specifies an optional `co` argument |
| `coroutine.running()` (5.1) | **New overload** — returns `thread?` only | In Lua 5.1, `coroutine.running()` returns `nil` when called from the main thread (no boolean) |
| `coroutine.running()` (≥5.2) | Fixed return type from `thread, string` → `thread, boolean` | The second return value is a boolean, not a string |

### 2. `debug.lua` (+42 / −18)

| Item | Change | Rationale |
|---|---|---|
| `debug.gethook()` | Return type corrected from `thread` → `function?, string, integer`; overload updated similarly | The function returns the hook function, mask string, and count — not a thread |
| `debuglib.InfoWhat` alias | Fixed typo: `lalinedefined` → `lastlinedefined` | Typo fix |
| `debug.getupvalue()` | Return types changed to `string?` and `any?` (nilable) | Returns `nil` when the index is out of range |
| `debug.setlocal()` | Return type changed to `string?` | Returns `nil` on invalid level |
| `debug.setupvalue()` | Return type changed to `string?` | Returns `nil` when index is out of range |
| `debug.setuservalue()` | Return type changed to `userdata?` | Can return `nil` per the manual |
| `debug.traceback()` | Split into **5.1** and **≥5.2** versions with distinct overloads; ≥5.2 version adds generic `T` passthrough for non-string messages | In Lua ≥5.2, if `message` is neither a string nor nil, it is returned as-is (identity passthrough). The previous annotation did not model this |

### 3. `global.lua` (+3 / −2)

| Item | Change | Rationale |
|---|---|---|
| `dofile()` | Return type changed from `table` → `any ...` | `dofile` returns all values returned by the chunk, not necessarily a table |
| `rawset()` | Added `@return table` | `rawset` returns the table being modified |
| `warn()` | Parameter renamed `message` → `msg1, ...`; accepts variadic strings | Lua 5.4 `warn` accepts multiple string fragments concatenated together |

### 4. `math.lua` (+20 / −2)

| Item | Change | Rationale |
|---|---|---|
| `math.abs()` | Doc-comment updated: "absolute value" → "maximum value between `x` and `-x`" | Matches the exact wording in the Lua 5.4 manual |
| `math.randomseed()` (5.1–5.3) | Added `@version 5.1, 5.2, 5.3` guard | Restricts the single-arg form to the correct versions |
| `math.randomseed()` (≥5.4) | **New overload** — accepts optional `x, y` and returns `integer, integer` | Lua 5.4 changed `randomseed` to accept two optional seed components and return them |
| `math.type()` | Return type fixed from `'integer'\|'float'\|'nil'` → `'integer'\|'float'\|nil` | `'nil'` (string literal) is incorrect; the function returns actual `nil` |

### 5. `os.lua` (+2 / −4)

| Item | Change | Rationale |
|---|---|---|
| `os.exit()` (≥5.2) | `code` parameter type changed to `boolean\|integer`, made optional; removed erroneous `@return integer` | `os.exit` does not return; `code` can be a boolean (`true`/`false`) in Lua ≥5.2 |
| `os.exit()` (5.1) | `code` made optional; removed `@return integer` | Same — `os.exit` never returns |

### 6. `string.lua` (+8 / −5)

| Item | Change | Rationale |
|---|---|---|
| `string.gmatch()` (≥5.4) | **New overload** with optional `init` parameter | Lua 5.4 added an `init` position argument |
| `string.pack()` | Parameter types widened from `string` → `string\|number\|integer` | `string.pack` packs numbers and integers, not just strings |

### 7. `table.lua` (+1 / −3)

| Item | Change | Rationale |
|---|---|---|
| `table.insert()` | Removed `@return integer` from both the overload and the main signature | `table.insert` returns nothing |
| `table.sort()` | Removed `@return integer` | `table.sort` returns nothing (sorts in-place) |

### 8. `utf8.lua` (+24 / −2)

| Item | Change | Rationale |
|---|---|---|
| `utf8.codes()` | Return type corrected from `string` → `fun(s: string, i?: integer): integer, integer` (iterator) | `utf8.codes` returns an iterator function, not a string |
| `utf8.codes()` (≥5.4) | **New overload** with `lax` parameter | Lua 5.4 added a `lax` boolean flag |
| `utf8.codepoint()` (≥5.4) | **New overload** with `lax` parameter | Same as above |
| `utf8.len()` | Split into base (≤5.3) and ≥5.4 versions; ≥5.4 adds `lax` parameter | Lua 5.4 added the `lax` flag |

---

### DIFF
--- a/crates/emmylua_code_analysis/resources/std/math.lua
+++ b/crates/emmylua_code_analysis/resources/std/math.lua
@@ -167,7 +167,7 @@ function math.random(m, n) end
 ---@param x integer
 function math.randomseed(x) end
 
----@version > 5.3
+---@version > 5.4
 ---
 --- When called with at least one argument, the integer parameters `x` and `y`
 --- are joined into a 128-bit seed that is used to reinitialize the pseudo-random


crates/emmylua_code_analysis/resources/std/debug.lua
```diff
@@ -289,22 +291,40 @@ function debug.setupvalue(f, up, value) end
...
 
---- Generates a traceback of the call stack.
---- When called with no arguments, returns a traceback of the current thread.
---- When the first argument is a thread, the traceback is generated for that thread;
---- the optional second argument (if a string) is prepended to the traceback, and the
---- optional third argument sets the level where the traceback should start (default is 1).
---- When the first argument is not a thread and not nil, it is treated as an optional message.
---- In that case, the traceback is generated for the current thread and the second argument,
---- if provided, specifies the starting level.
+---@version 5.1
```
CppCXY: Add JIT

crates/emmylua_code_analysis/resources/std/global.lua
```diff
@@ -469,8 +470,9 @@ function unpack(list, i, j) end
 
 ---@version > 5.4
 ---
----@param message string
-function warn(message) end
+---@param msg1 string
+---@param ... string
+function warn(msg1, ...) end
```
CppCXY, line 481: warn first param support this: 
///   @on    - enable warnings (stderr mode)
///   @off   - disable warnings
///   @store - store warnings in _WARN global
///   @normal - restore normal stderr output

EfveZombie: `warn` behaviors may changed by `lua_setwarnf`, it only checks the arguments type:

```lua
static int luaB_warn (lua_State *L) {
  int n = lua_gettop(L);  /* number of arguments */
  int i;
  luaL_checkstring(L, 1);  /* at least one argument */
  for (i = 2; i <= n; i++)
    luaL_checkstring(L, i);  /* make sure all arguments are strings */
  for (i = 1; i < n; i++)  /* compose warning */
    lua_warning(L, lua_tostring(L, i), 1);
  lua_warning(L, lua_tostring(L, n), 0);  /* close warning */
  return 0;
}
```

Some descriptions should be enough?

crates/emmylua_code_analysis/resources/std/math.lua
```diff
@@ -160,12 +160,30 @@ function math.rad(x) end
...
 function math.random(m, n) end
 
+---@version 5.1, 5.2, 5.3
 ---
 --- Sets `x` as the "seed" for the pseudo-random generator: equal seeds
 --- produce equal sequences of numbers.
 ---@param x integer
 function math.randomseed(x) end
 
+---@version > 5.3
```
CppCXY: The '>' here actually means 'greater than or equal to', because that's how luals works.

crates/emmylua_code_analysis/resources/std/string.lua
```diff
@@ -230,9 +237,9 @@ function string.match(s, pattern, init) end
 --- Returns a binary string containing the values `v1`, `v2`, etc. packed (that
 --- is, serialized in binary form) according to the format string `fmt`.
 ---@param fmt string
----@param v1 string
----@param v2? string
----@param ... string
+---@param v1 string|number|integer
```
CppCXY, line 240: If the parameter is number, then it includes integer.

EfveZombie: I did this on purpose, `number|integer` (or `integer|number`)  shows `string.pack` accepts 3 types of argument and it may behave differently with `integer` & `number`

PR #974, opened 2026-03-02
feat(code-analysis): support @return_overload flow narrowing

lewis6991:
This PR adds correlated `---@return_overload` rows for multi-return APIs.

Today, plain `---@return` docs flatten each return slot independently, which loses the relationship between values
like `ok` and `result`. `---@return_overload` lets each annotation row describe one possible return tuple, so the
analyzer can preserve that correlation through inference and control-flow narrowing.

```lua
---@return_overload true, T
---@return_overload false, string
local ok, result = pcall(foo)

if not ok then
  error(result)
end

-- `result` narrows to `T` here
```

## What Changed

- Parse and document `---@return_overload`.
- Store return overload rows separately from plain return docs and surface them in tooling output.
- Use overload rows during return inference, including generic and variadic tails.
- Pad missing trailing overload slots with nil so short rows still produce correct tuple shapes.
- Narrow correlated return slots on truthy/falsy and literal-equality checks while staying sound across reassignments,
  mixed RHS calls, and branch joins.
- Annotate stdlib pcall with correlated return rows so status/result flows narrow correctly.
- Update incomplete-signature diagnostics to understand return overload rows.

## Coverage Added

- Parser/doc tests for the new tag.
- Analyzer tests for generic and variadic return rows.
- Flow tests for correlated narrowing, reassignments, joins, and equality guards.
- pcall regression coverage.
- Hover, semantic-token, and diagnostic coverage for the new annotation.


lewis6991:
@CppCXY what do you think about this?

Admittedly I got AI to implement all of this, but I guided it best I could and reviewed all the changes.

I'm not sure how good the overall strategy is here, but it seems to align with how other stuff is analysed. There's some slight code duplication with literal comparison narrowing, but that can be improved later.

CppCXY:
How do return overloads match with generics, and what would `pcall(pcall, f)` return?

lewis6991:
Good catch, It didn't handle it well, though either did `main`. I've made some changes to fix this.

I ended up running a "review, fix, minimize" loop in codex which ended up finding lots of edge cases to fix. This has made this PR much bigger, but ~40% is tests.

On `main`:

```lua
local ok, status, payload = pcall(pcall, produce)
-- ok - boolean
-- status - boolean|string
-- payload - R1|string
```

This PR (with and without `@return_overload`):

```lua
local ok, status, payload = pcall(pcall, produce)
-- ok - true|false
-- status - string|true|false
-- payload - string|integer
```

crates/emmylua_parser/src/grammar/doc/tag.rs
```diff
@@ -38,7 +38,7 @@ fn parse_tag_detail(p: &mut LuaDocParser) -> DocParseResult {
         LuaTokenKind::TkTagField => parse_tag_field(p),
         LuaTokenKind::TkTagType => parse_tag_type(p),
         LuaTokenKind::TkTagParam => parse_tag_param(p),
-        LuaTokenKind::TkTagReturn => parse_tag_return(p),
+        LuaTokenKind::TkTagReturn | LuaTokenKind::TkTagReturnOverload => parse_tag_return(p),
```
CppCXY: You should use different function instead of sharing parse_tag_return, or design a new syntax for `@return` like (true, T...) | (false, string).
```lua
---@return (true, R1...)
---| (false, string)
```

xuhuanzy: 这个可以用条件泛型替代

xuhuanzy:
You might consider using conditional generics to handle this instead of introducing new syntax.

CppCXY:
> You might consider using conditional generics to handle this instead of introducing new syntax.

I think we should stick to simple rules. The current complex generics are immature and need another round of refactoring and rethinking. We can adopt `@return_overload`, but I don't want to have to write `@return` again after using `@return_overload`.

xuhuanzy:
After analysis, for conditional generics to work as intended, built-in generics would have to be introduced. In that case, there’s essentially no fundamental difference from `@return_overload`, so using `@return_overload` is a viable approach.

crates/emmylua_parser/src/grammar/doc/tag.rs
CppCXY: Besides using different parse functions, we should also use different structs — I don't really want them to share the same Struct. We also need to consider how hover will display related information due to the introduction of return_overload.

lewis6991: Noted. Originally these were completely separate, but a minimize pass joined them. I'll make them separate again and look into hover.

lewis6991: > or design a new syntax for @return like (true, T...) | (false, string).

I missed this part of the comment. I'll give this a go.

lewis6991:
I've explored several options for alternative syntax. Here is a summary. What are people thoughts?

---

| Syntax | Verdict | Notes |
| --- | --- | --- |
| `---@return_overload true, T` | Baseline | Already implemented. Most explicit, but adds a new tag. |
| `---@return true, T` then `---@return false, E` | No | Repeated `@return` lines already mean more return slots, not alternative rows. |
| `---@return ty1, ty2 \| ty3, ty4` | No | Parses as slot 1 = `ty1`, slot 2 = `ty2\|ty3`, slot 3 = `ty4`. |
| `---@return [true, T] \| [false, E]` | Viable | Reuses existing tuple + union syntax. Downside: `[]` does not feel ideal for return rows. |
| `---@return (true, T) \| (false, E)` | Expensive | Nice syntax, but `()` currently mean grouping a single type, not tuple rows. |
| `---@return true, T; false, E` | Viable | Clean row separator, but single-line compactness is not needed. |
| `---@return` + `---; true, T` | Viable | Good multiline partner for `;`. Explicit row separator. |
| `---@return` + `---\| true, T` | Possible, not preferred | `---\|` already reads like union or enum continuation, so the notation is confusing. |
| `---@return[fail] false, E` | Viable | Gives named rows. Downside: `[]` already carry tuple meaning elsewhere. |
| `---@return fail: false, E` | Viable | Readable named-row form. Requires `@return` to recognize a row label prefix. |
| `---@return : false, E` | Strong candidate | Small, explicit row marker. No new tag, no `[]`, no overloading `\|`. |

## Current Parser Constraints

- `@return` currently parses as comma-separated return slots.
- `|` already means type union, so it is a poor row separator.
- `[]` are the existing tuple syntax.
- `()` currently mean grouping a single type, not tuple rows.
- `---|` is already used as a multiline union or enum continuation marker.

## Current Preference

### Option A: Empty Row Marker

```lua
---@return boolean
---@return T|E
---@return : true, T
---@return : false, E
```

Pros:

- explicit
- local change to `@return`
- no new tag
- no visual conflict with `---|`

### Option B: Named Rows

```lua
---@return boolean
---@return T|E
---@return ok: true, T
---@return fail: false, E
```

Pros:

- good readability in docs and hover
- rows have stable names
- still avoids a new tag

Tradeoff:

- needs a decision on whether unlabeled `:` rows and labeled rows should both be supported

### Option C: Multiline Row Separator

```lua
---@return
---; true, T
---; false, E
```

Pros:

- explicit multiline structure
- pairs naturally with `;` if we ever want a single-line form later

Tradeoff:

- more notation than `@return : ...`

CppCXY:
All approaches based on `@return` have a problem: because multiple `@return` tags are allowed, the semantics become completely ambiguous when more than one `@return` appears. I suggest keeping return_overload unchanged.

crates/emmylua_parser/src/grammar/doc/test.rs
```diff
@@ -132,6 +132,29 @@ Syntax(Chunk)@0..163
...
     }
 
+    #[test]
+    fn test_return_overload_tag() {
+        let code = r#"
+        ---@return_overload true, integer
+        "#;
+        let tree = LuaParser::parse(code, ParserConfig::default());
+        let result = format!("{:#?}", tree.get_red_root());
+        assert!(result.contains("Syntax(DocTagReturnOverload)"));
```
CppCXY: The AI tends to validate generated strings in this way, but that check ignores the hierarchical relationships between nodes and whether there are misparsed nodes. I suggest validating the parse tree using the existing method — we usually call `print_ast` to print the tree and then just copy it over.

crates/emmylua_parser/src/grammar/doc/tag.rs
```diff
@@ -369,6 +370,25 @@ fn parse_tag_return(p: &mut LuaDocParser) -> DocParseResult {
...
 }
 
+// ---@return_overload true, integer
+// ---@return_overload false, string
+fn parse_tag_return_overload(p: &mut LuaDocParser) -> DocParseResult {
+    p.set_lexer_state(LuaDocLexerState::Normal);
+    let m = p.mark(LuaSyntaxKind::DocTagReturnOverload);
+    p.bump();
+
+    parse_type(p)?;
```
CppCXY, line 380: Should the first parameter of return_overload be limited to true|false, or should it be restricted to literal constants? Does the current return_overload support all literal types?

lewis6991: As far as I'm aware I think it supports any types as it has correlated type tracking and uses the normal narrowing functions.

EDIT: just double-checked, it doesn't support any type, but does support any literal type.

lewis6991: Added a commit to make the type checks broader.

Issue #964, opened 2026-02-23
How to check "no-unknown" equivalent from lua-language-server - assert every variable/function is typed


dquixote-jr:
### Problem

The following file (test.lua) returns no errors when doing a check with `emmylua_check -c ./.emmyrc.json --warnings-as-errors test.lua`:

```lua
local function do_add(x, y)
  return x + y
end

return do_add
```

My `.emmyrc.json` is as follows:

```json
{
  "format": {
    "externalTool": {
      "program": "stylua",
      "args": ["-", "--stdin-filepath", "${file}"]
    }
  },
  "runtime": {
    "version": "LuaJIT"
  },
  "diagnostics": {
    "globals": ["ngx"],
    "disable": ["unnecessary-if", "invert-if"]
  }
}
```

### Expected behaviour

That emmylua_check should return an error when there is a function which is not typed.

### Notes

I'm trying to make sure everything is typed properly in my code base but I'm not seeing any way to do that for functions. Is there any way or workaround to do this right now?

CppCXY:
maybe only this one: `incomplete-signature-doc`
```json
  "diagnostics": {
    "enables": [
      "incomplete-signature-doc"
    ]
  },
```

dquixote-jr:
The diagnostic you mentioned, `"incomplete-signature-doc"`, was not working properly for local functions. I've made a PR to fix that.

PR #975, opened 2026-03-04
feat: update incomplete-signature-doc to also error on non-global functions

dquixote-jr:
relates - https://github.com/EmmyLuaLs/emmylua-analyzer-rust/issues/964

### DIFF
--- a/crates/emmylua_code_analysis/src/diagnostic/test/incomplete_signature_doc_test.rs
+++ b/crates/emmylua_code_analysis/src/diagnostic/test/incomplete_signature_doc_test.rs
@@ -24,11 +24,29 @@ mod tests {
         let mut ws = VirtualWorkspace::new();
         ws.enable_full_diagnostic();
 
+        // TODO: closures are not detected as functions in the semantic model
+        // assert!(!ws.check_code_for(
+        //     DiagnosticCode::MissingGlobalDoc,
+        //     r#"
+        //     local c = function(x, y)
+        //         return x + y
+        //     end
+        //     "#
+        // ));
+
         assert!(!ws.check_code_for(
             DiagnosticCode::IncompleteSignatureDoc,
             r#"
             local function do_add(x, y)
                 return x + y
+            end
+            "#
+        ));
+
+        assert!(!ws.check_code_for(
+            DiagnosticCode::IncompleteSignatureDoc,
+            r#"
+            local function noop()
             end
             "#
         ));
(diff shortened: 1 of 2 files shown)


crates/emmylua_code_analysis/src/diagnostic/test/incomplete_signature_doc_test.rs
```diff
@@ -27,9 +27,8 @@ mod tests {
...
             DiagnosticCode::IncompleteSignatureDoc,
             r#"
-            ---@param p number
-            local function FLPR3(p, e)
-                return 0
+            local function do_add(x, y)
+                return x + y
             end
             "#
         ));
```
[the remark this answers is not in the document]
dquixote-jr: No, the ws.check_code_for returns false on the error being in the raised diagnostics.

CppCXY:
Please fix the check

dquixote-jr:
done

xuhuanzy:
@CppCXY 需要考虑对闭包跟一些简单函数的影响，先别合

xuhuanzy:
If we want to extend the diagnosis to be effective for non-global variables, then we need to handle some simple cases.
```lua
local function a() -- No function parameters should trigger a warning.

end

function b(a) -- issue warning

end

local c = function(a) -- A warning should be issued.

end
```

dquixote-jr:
Added tests for the first 2, part 3 is broken right now as well, it doesn't work for global functions either, e.g. for the following:
```
c = function(a) -- A warning should be issued.
end
```

Type checker does not issue any warnings. So I'm not sure if fixing that is supposed to be part of this change. Should be a different PR.

crates/emmylua_code_analysis/src/diagnostic/test/incomplete_signature_doc_test.rs
```diff
@@ -24,12 +24,29 @@ mod tests {
         let mut ws = VirtualWorkspace::new();
         ws.enable_full_diagnostic();
 
+        // TODO: closures are not detected as functions in the semantic model
+        // assert!(!ws.check_code_for(
```
CppCXY, line 28: If you don't know how to handle it, you can ask AI for help.

dquixote-jr: Does it have to be this PR?
Closures are not detected for global functions either. So I'm not sure if fixing that needs to be done right now.
The functionality that I am adding is unrelated to fixing the function detection.

CppCXY: Ok, That sounds reasonable.

Issue #873, opened 2025-12-05
Unable to ignore files for diagnostics


clason:
Currently, files that are ignored (either via the `ignoreDirs` or `ignoreGlobs` setting) are not parsed at all (which was a deliberate change).

However, this is a problem if your project contains vendored library files that trigger diagnostics -- in this case, you _do_ want these files to be parsed but you simply don't want any diagnostics to be reported, especially with `emmylua_check`. (Otherwise you'll trigger errors in _your_ files because `emmylua` doesn't know about the vendored API and types.)

(I also didn't find a (documented) way of marking files as ignored via `--- emmylua: ignore` comments, but an **additional** config setting for "parse but don't diagnose" would be more useful.)

CppCXY:
If it is a file, adding meta will not trigger a diagnosis. If it is a directory added to the library, it will not be diagnosed either. Note that the library is resolved before the workspace, and the library is unaware of the definitions in the main workspace.

clason:
For me, it's a single (non-meta) file vendored inside the main sources (say, `lua/foo/vendored.lua` if my main module is `foo`), so adding it to the library will have it processed twice. Unless I'm missing something from your suggestion?

xuhuanzy:
`---@meta`鸟用没有, meta 应该只过滤一些诊断而不是过滤全部

CppCXY:
Currently the library only supports directories, but we should be able to improve this.

clason:
Adding `---@meta` to the top of the file works, for the record. It would be cleaner not having to touch vendored files at all (especially if there are multiple), though, and this annotation breaks luals (which is an issue when you want or need to migrate incrementally, like I do).

clason:
A related issue around the two different aspects ignoring files is the following: In my (Neovim-based) projects, I download the Neovim runtime (and several other dependencies) to a `.test-deps` folder. I am then specifying (via an environment variable `$VIMRUNTIME`) the runtime as a workspace library; at the same time, I do not want to check all the dependencies themselves. So my `luarc.json` has the following:
```json
  "workspace": {
    "library": [
      "$VIMRUNTIME",
      "${3rd}/busted/library"
    ],
    "ignoreDir": [
      ".test-deps",
      "tests"
    ]
  },
```
[...]

PR #981, opened 2026-03-09
fix: honor explicit library roots despite workspace excludes

lewis6991:
Explicitly configured libraries and package directories should still be
indexed even when they live under paths excluded from the main
workspace.

Separate the expanded file-matching roots from the client workspace
roots so explicit library roots can bypass parent ignoreDir filtering
during collection and watched-file matching without being fed back into
init_analysis as main workspaces on reload.

Keep global ignore globs and library-specific ignore rules applying
inside those library roots, and add regression coverage for matching,
collection, and reload state.


clason:
Note that this was already fixed back on December 8 but got reverted in https://github.com/EmmyLuaLs/emmylua-analyzer-rust/commit/faf2233eb467245b73e08fc2dd3258423241024c (see #873).

clason:
This fixes my issue.

PR #984, opened 2026-03-11
fix: refresh workspace match and watch state on reload

lewis6991:
## Summary

Config reload rebuilt analysis state, but the language server could keep using stale workspace matching state and stale watched-file registrations. In practice, changes to `workspace.library`, `workspace.packageDirs`, or `workspace.workspaceRoots` were not always reflected until restart, removed roots could stay indexed, and external library roots could keep outdated watch coverage.

## What changed

- rebuild the effective workspace folder set from:
  - client workspace roots
  - configured `workspaceRoots`
  - configured `library`
  - configured `packageDirs`
- refresh workspace matching state whenever config is reloaded
- re-register watched-file globs after reload
- rebuild notify watchers for roots outside the editor workspace
- clear non-stdlib workspaces before reloading files so removed roots drop out of analysis
- preserve unsaved open-buffer text across reloads when files enter or leave the workspace
- clear push diagnostics for files removed by reload
- separate debounced config reloads from reindex requests so a save no longer cancels a pending config reload
- share the workspace file matcher between `emmylua_code_analysis` and `emmylua_ls` to remove duplicated matching logic

## Why

Before this change:

- adding a library or workspace root via config could require a restart before the LS treated those files as workspace files
- removing a library could leave files indexed and diagnostics uncleared
- external library roots could keep stale watcher state after reload
- reindex scheduling could interfere with config reload scheduling
- unsaved open-file text could be lost across workspace membership changes during reloadnalysis and LS**


crates/emmylua_ls/src/handlers/initialized/mod.rs
```diff
@@ -99,58 +92,28 @@ pub async fn initialized_handler(
...
     lsp_features: &LspFeatures,
     workspace_folders: Vec<WorkspaceFolder>,
     emmyrc: Arc<Emmyrc>,
+    open_files: Vec<(lsp_types::Uri, String)>,
 ) {
     let mut mut_analysis = analysis.write().await;
 
     // update config
     mut_analysis.update_config(emmyrc.clone());
+    mut_analysis.clear_non_std_workspaces();
```
CppCXY: this will clean the remote file, This may cause repeatedly download the schema.json file.

PR #989, opened 2026-03-20
fix(flow): apply deferred narrows after antecedent resolution

lewis6991:
When walking backward through flow, collect condition narrows as pending
actions instead of applying them while recursively querying antecedent
types. The eager path mixed narrowing with antecedent resolution, so
stacked guards could re-enter flow inference from inside condition
evaluation and build deep recursive chains across repeated truthiness,
type-guard, signature-cast, and correlated return-overload checks.

Resolve the antecedent type first, then apply the pending narrows in
reverse order. That keeps narrowing as a post-pass over an
already-resolved input, avoids re-entering the same condition chain
while answering the current flow query, and lets same-variable
self/member guards wait until earlier guards have narrowed the receiver
enough for reliable lookup.

Key the flow cache by whether condition narrowing is enabled, and
separate assignment source lookup from condition application. Reuse a
narrowed source only when the RHS preserves that precision; broader RHS
expressions fall back to the antecedent type with condition narrowing
disabled so reassignment clears stale branch narrows, while exact
literals and compatible partial table/object rewrites still preserve
useful narrowing.

Add regression coverage for stacked guards, correlated overload joins,
pcall aliases, and assign/return diagnostics.

---

These added tests fail on `main`:
- test_stacked_same_var_type_guards_build_semantic_model
- test_stacked_same_field_truthiness_guards_build_semantic_model
- test_stacked_return_cast_guards_build_semantic_model
- test_stacked_return_cast_self_guards_build_semantic_model
- test_assignment_from_wider_single_return_call_drops_branch_narrowing
- test_assignment_after_pending_return_cast_guard_drops_branch_narrowing
- test_assignment_after_binary_call_guard_eq_false_drops_branch_narrowing
- test_assignment_after_mixed_eager_and_pending_guards_drops_branch_narrowing
- test_assignment_from_nullable_union_keeps_rhs_members
- test_assignment_from_partially_overlapping_union_keeps_rhs_members
- test_assignment_from_broad_string_drops_literal_branch_narrowing
- test_partial_table_reassignment_with_conflicting_discriminant_drops_branch_narrowing
- test_pcall_stacked_alias_guards_build_semantic_model
- test_return_overload_stacked_eq_guards_build_semantic_model
- test_return_overload_stacked_mixed_guards_build_semantic_model
- test_return_overload_uncorrelated_later_guard_keeps_prior_narrowing
- test_return_overload_unmatched_target_call_keeps_guard_union
- test_return_overload_unmatched_target_root_then_truthiness_guard
- test_return_overload_unmatched_target_root_then_type_guard
- test_return_overload_post_guard_reassign_clears_mixed_root_narrowing
- test_return_overload_branch_reassign_to_different_call_preserves_matching_root_narrowing
- test_return_overload_branch_reassign_to_different_call_narrows_alternate_matching_root

### DIFF
--- a/crates/emmylua_code_analysis/src/semantic/infer/narrow/condition_flow/binary_flow.rs
+++ b/crates/emmylua_code_analysis/src/semantic/infer/narrow/condition_flow/binary_flow.rs
@@ -289,7 +289,7 @@ fn maybe_type_guard_binary_action(
     };
 
     if maybe_var_ref_id == *var_ref_id {
-        return Ok(Some(ConditionFlowAction::Pending(
+        return Ok(Some(ConditionFlowAction::pending(
             PendingConditionNarrow::TypeGuard {
                 narrow,
                 condition_flow,
@@ -331,7 +331,7 @@ fn maybe_type_guard_binary_action(
         &narrowed_discriminant_type,
     )?
     .map(PendingConditionNarrow::Correlated)
-    .map(ConditionFlowAction::Pending))
+    .map(ConditionFlowAction::pending))
 }
 
 /// Maps the string result of Lua's builtin `type()` call to the corresponding `LuaType`.
@@ -441,7 +441,7 @@ fn get_var_eq_condition_action(
                     &narrowed_discriminant_type,
                 )?
                 .map(PendingConditionNarrow::Correlated)
-                .map(ConditionFlowAction::Pending)
+                .map(ConditionFlowAction::pending)
                 .unwrap_or(ConditionFlowAction::Continue));
             }
 
@@ -452,14 +452,14 @@ fn get_var_eq_condition_action(
                     if var_ref_id.is_self_ref() && !right_expr_type.is_nil() {
                         TypeOps::Remove.apply(db, &right_expr_type, &LuaType::Nil)
                     } else {
-                        return Ok(ConditionFlowAction::Pending(PendingConditionNarrow::Eq {
+                        return Ok(ConditionFlowAction::pending(PendingConditionNarrow::Eq {
                             right_expr_type,
                             condition_flow,
                         }));
                     }
                 }
                 InferConditionFlow::FalseCondition => {
-                    return Ok(ConditionFlowAction::Pending(PendingConditionNarrow::Eq {
+                    return Ok(ConditionFlowAction::pending(PendingConditionNarrow::Eq {
                         right_expr_type,
                         condition_flow,
                     }));
@@ -499,7 +499,7 @@ fn get_var_eq_condition_action(
 
             let right_expr_type = infer_expr(db, cache, right_expr)?;
             if condition_flow.is_false() {
-                return Ok(ConditionFlowAction::Pending(PendingConditionNarrow::Eq {
+                return Ok(ConditionFlowAction::pending(PendingConditionNarrow::Eq {
                     right_expr_type,
                     condition_flow,
                 }));
--- a/crates/emmylua_code_analysis/src/semantic/infer/narrow/condition_flow/correlated_flow.rs
+++ b/crates/emmylua_code_analysis/src/semantic/infer/narrow/condition_flow/correlated_flow.rs
@@ -4,7 +4,7 @@ use emmylua_parser::{LuaAstPtr, LuaCallExpr, LuaChunk};
 
 use crate::{
     DbIndex, FlowId, FlowNode, FlowTree, InferFailReason, LuaDeclId, LuaFunctionType,
-    LuaInferCache, LuaType, TypeOps, infer_expr, instantiate_func_generic,
+    LuaInferCache, LuaSignature, LuaType, TypeOps, infer_expr, instantiate_func_generic,
     semantic::infer::{
         VarRefId,
         narrow::{get_single_antecedent, get_type_at_flow::get_type_at_flow},
@@ -23,25 +23,33 @@ struct SearchRootCorrelatedTypes {
     deferred_known_call_target_types: Option<Vec<LuaType>>,
 }
 
+#[derive(Debug)]
+struct CollectedCorrelatedTypes {
+    matching_target_types: Vec<LuaType>,
+    correlated_candidate_types: Vec<LuaType>,
+    unmatched_target_types: Vec<LuaType>,
+    has_unmatched_discriminant_origin: bool,
+    has_opaque_target_origin: bool,
+}
+
 impl CorrelatedConditionNarrowing {
     pub(in crate::semantic::infer::narrow) fn apply(
-        self,
+        &self,
         db: &DbIndex,
         antecedent_type: LuaType,
     ) -> LuaType {
         let mut root_target_types = Vec::new();
         let mut found_matching_root = false;
-        for root_types in self.search_root_correlated_types {
-            let SearchRootCorrelatedTypes {
-                matching_target_types,
-                mut uncorrelated_target_types,
-                deferred_known_call_target_types,
-            } = root_types;
+        for root_types in &self.search_root_correlated_types {
+            let matching_target_types = &root_types.matching_target_types;
+            let mut uncorrelated_target_types = root_types.uncorrelated_target_types.clone();
+            let deferred_known_call_target_types =
+                root_types.deferred_known_call_target_types.as_deref();
 
             let root_matching_target_type = if matching_target_types.is_empty() {
                 None
             } else {
-                let matching_target_type = LuaType::from_vec(matching_target_types);
+                let matching_target_type = LuaType::from_vec(matching_target_types.clone());
                 let narrowed_correlated_type =
                     TypeOps::Intersect.apply(db, &antecedent_type, &matching_target_type);
                 if narrowed_correlated_type.is_never() {
@@ -185,13 +193,13 @@ fn collect_correlated_types_from_search_root(
         condition_position,
         search_root_flow_id,
     );
-    let (
-        root_matching_target_types,
-        root_correlated_candidate_types,
-        root_unmatched_target_types,
+    let CollectedCorrelatedTypes {
+        matching_target_types: root_matching_target_types,
+        correlated_candidate_types: root_correlated_candidate_types,
+        unmatched_target_types: root_unmatched_target_types,
         has_unmatched_discriminant_origin,
         has_opaque_target_origin,
-    ) = collect_matching_correlated_types(
+    } = collect_matching_correlated_types(
         db,
         cache,
         root,
@@ -277,7 +285,7 @@ fn collect_matching_correlated_types(
     discriminant_refs: &[crate::DeclMultiReturnRef],
     target_refs: &[crate::DeclMultiReturnRef],
     narrowed_discriminant_type: &LuaType,
-) -> Result<(Vec<LuaType>, Vec<LuaType>, Vec<LuaType>, bool, bool), InferFailReason> {
+) -> Result<CollectedCorrelatedTypes, InferFailReason> {
     let mut matching_target_types = Vec::new();
     let mut correlated_candidate_types = Vec::new();
     let mut unmatched_target_types = Vec::new();
@@ -304,18 +312,16 @@ fn collect_matching_correlated_types(
             correlated_discriminant_call_expr_ids.insert(discriminant_call_expr_id);
             correlated_target_call_expr_ids.insert(target_ref.call_expr.get_syntax_id());
             correlated_candidate_types.extend(overload_rows.iter().map(|overload| {
-                crate::LuaSignature::get_overload_row_slot(overload, target_ref.return_index)
+                LuaSignature::get_overload_row_slot(overload, target_ref.return_index)
             }));
             matching_target_types.extend(overload_rows.iter().filter_map(|overload| {
-                let discriminant_type = crate::LuaSignature::get_overload_row_slot(
-                    overload,
-                    discriminant_ref.return_index,
-                );
+                let discriminant_type =
+                    LuaSignature::get_overload_row_slot(overload, discriminant_ref.return_index);
                 if !TypeOps::Intersect
                     .apply(db, &discriminant_type, narrowed_discriminant_type)
                     .is_never()
                 {
-                    return Some(crate::LuaSignature::get_overload_row_slot(
+                    return Some(LuaSignature::get_overload_row_slot(
                         overload,
                         target_ref.return_index,
                     ));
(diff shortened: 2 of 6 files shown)


clason:
regression after the last push:
```
--- lua/nvim-treesitter/config.lua [3 warnings]
warning: expected `table<any,string>` but found `(string[]|string|(string))?`.  [param-type-mismatch]
  -->: lua/nvim-treesitter/config.lua:124:7

  123 │       end,
  124 │       languages
  125 │     )
```

**EDIT:** fixed.

crates/emmylua_code_analysis/src/semantic/infer/narrow/condition_flow/binary_flow.rs
```diff
@@ -235,7 +238,7 @@ fn maybe_type_guard_binary(
     left_expr: LuaExpr,
     right_expr: LuaExpr,
     condition_flow: InferConditionFlow,
-) -> Result<ResultTypeOrContinue, InferFailReason> {
+) -> Result<Option<ConditionFlowAction>, InferFailReason> {
```
CppCXY, line 241: Using "option" here instead of "continue" somewhat reduces readability.

crates/emmylua_code_analysis/src/semantic/infer/narrow/condition_flow/index_flow.rs
```diff
@@ -16,108 +14,43 @@ use crate::{
 #[allow(clippy::too_many_arguments)]
 pub fn get_type_at_index_expr(
     db: &DbIndex,
-    tree: &FlowTree,
+    _tree: &FlowTree,
```
CppCXY: Remove unused parameters directly, just like you did earlier.

crates/emmylua_code_analysis/src/semantic/infer/narrow/condition_flow/binary_flow.rs
lewis6991: We use option here because `try_get_at_eq_or_neq_expr` has:

```rust
    if let Some(action) = maybe_type_guard_binary_action(
        db,
        tree,
        cache,
        root,
        var_ref_id,
        flow_node,
        left_expr.clone(),
        right_expr.clone(),
        condition_flow,
    )? {
        return Ok(action);
    }
```

Using continue would make this a little worse I think.

crates/emmylua_code_analysis/src/semantic/infer/narrow/mod.rs
```diff
@@ -71,22 +71,11 @@ fn get_var_ref_type(db: &DbIndex, cache: &mut LuaInferCache, var_ref_id: &VarRef
     }
 }
 
-fn get_single_antecedent(tree: &FlowTree, flow: &FlowNode) -> Result<FlowId, InferFailReason> {
+fn get_single_antecedent(_tree: &FlowTree, flow: &FlowNode) -> Result<FlowId, InferFailReason> {
```
CppCXY: also remove, Many places no longer use FlowTree and use only FlowNode — does that mean they've abandoned having multiple decision paths?

lewis6991: Removed.

The multiple-path flow model is still intact. `FlowAntecedent::Multiple` is still represented in the CFG, expanded via `get_multi_antecedents`, and used at branch joins in `get_type_at_flow`. `get_single_antecedent` is only the linear-predecessor helper, so it intentionally rejects `Multiple`.

CppCXY:
Is there any performance degradation or improvement when processing a single large file—for example, a definition file with many assignment statements of about 10k lines?

lewis6991:
The motivation for this change was to improve performance by reducing recursion. The return_overload patch pushed the current flow analysis to its limits and I found a case where it caused indexing to hang.

To be extra sure, I'll run this on a few projects and get back to you.

clason:
...and also to fix a regression that prevents processing certain (i.e., my) projects ;) So the mentioned hang is not hypothetical.

CppCXY:
> ...and also to fix a regression that prevents processing certain (i.e., my) projects ;) So the mentioned hang is not hypothetical.

I often receive reports of performance issues from the language server on large configuration tables and on config files with many assignment definitions; I’m aware of it, but I haven’t gotten around to fixing it.

clason:
No, this is not about a performance issue. This is a very specific regression introduced in #985 which makes analysis hang at 100% CPU for hours on nvim-treesitter. This PR fixes that regression so analysis completes quickly again.

lewis6991:
> > ...and also to fix a regression that prevents processing certain (i.e., my) projects ;) So the mentioned hang is not hypothetical.
> 
> I often receive reports of performance issues from the language server on large configuration tables and on config files with many assignment definitions; I’m aware of it, but I haven’t gotten around to fixing it.

I did add a stress regression test for a large linear assignment-heavy file, `test_large_linear_assignment_file_builds_semantic_model`, so CI now covers that shape as well. It is not a benchmark, but it should catch this refactor regressing back into pathological flow walking on long assignment chains.

lewis6991:
I've tested indexing neovim, gitsigns and they index very quickly. I also tried my neovim config which includes all my plugins (~4000 files) and that indexed reasonably quickly too (3-6 seconds).

xuhuanzy:
The performance issues come from game development projects, where many projects have over ten thousand files and individual files are megabytes in size.
This is completely different from your usage environment, and I think it's hard for you to imagine the specific tests

lewis6991:
Ok, then can you help me confirm whether this PR improves or decreases performance on such projects? I'm yet to see a project referenced in any issue.

CppCXY:
[lua-bench.zip](https://github.com/user-attachments/files/26272112/lua-bench.zip)
 as a example file, By comparing load times with the latest release on GitHub, I found that it is noticeably twice as slow in diagnostics, while loading performance is probably about the same. This is likely the cumulative result of several PRs, although most of those PRs were submitted by you.

CppCXY:
and I found the collectgarbage will become unknown at line: 9249
image
However, it probably has nothing to do with you, because this issue existed in earlier versions.

lewis6991:
> [lua-bench.zip](https://github.com/user-attachments/files/26272112/lua-bench.zip) as a example file, By comparing load times with the latest release on GitHub, I found that it is noticeably twice as slow in diagnostics, while loading performance is probably about the same. This is likely the cumulative result of several PRs, although most of those PRs were submitted by you.

Thanks for this. We should probably add this to CI in some form so we can have some better control on performance regressions.

The last few commits do improve analysis quite a bit, and that obviously come at some cost, though now we have a reference project that can use help guide optimizations.

I'll check how this PR specially affects this workload. If it is the same or improves things then I hope we can merge this and add further optimizations in follow ups.

lewis6991:
[lua-bench.zip](https://github.com/user-attachments/files/26272112/lua-bench.zip) does the same recursion hang on `main`. With this PR diagnostics complete in about 2.5s on my machine using emmylua_check.

lewis6991:
Latest release: 1.38s
This PR: 2.56s

So it does slow things down quite a bit. I'll try and work on optimizing that.

lewis6991:
Added some caching.

  Ran a quick local `lua-bench` comparison between current HEAD (`60798f5c`) and the latest release tag `0.21.0` (`1546f82e`) using `emmylua_check lua-
  bench` on `lua-bench/10k_row_code.lua`.

  Method:
  - rebuilt both `emmylua_check` binaries locally with the same toolchain
  - did 1 warm-up run, then 5 timed warm-cache runs for each build

  | build | runs (s) | average |
  | --- | --- | ---: |
  | `0.21.0` | `1.34, 1.32, 1.33, 1.34, 1.35` | `1.336s` |
  | `HEAD` | `0.50, 0.54, 0.54, 0.51, 0.54` | `0.526s` |


  On this workload, HEAD is about 2.5x faster than the release build.

xuhuanzy:
How is the memory usage? Has it increased significantly?
To be clear, game development is the top priority for this project, and we are very concerned about memory usage.

lewis6991:
I don't know. It's most likely going to be better then `main`.

CppCXY:
From the startup logs, it looks like the language server parses the workspace twice during startup.
```log
[2026-03-27 10:08:25 +08:00 INFO emmylua_code_analysis::profile] update files: cost 348.4692ms
[2026-03-27 10:08:25 +08:00 INFO emmylua_code_analysis::profile] module analyze: cost 3.2618ms
[2026-03-27 10:08:25 +08:00 INFO emmylua_code_analysis::profile] decl analyze: cost 394.4664ms
[2026-03-27 10:08:25 +08:00 INFO emmylua_code_analysis::profile] doc analyze: cost 75.0438ms
[2026-03-27 10:08:25 +08:00 INFO emmylua_code_analysis::profile] flow analyze: cost 102.8077ms
[2026-03-27 10:08:26 +08:00 INFO emmylua_code_analysis::profile] lua analyze: cost 787.0893ms
[2026-03-27 10:08:27 +08:00 INFO emmylua_code_analysis::profile] resolve analyze: cost 614.7488ms
[2026-03-27 10:08:27 +08:00 INFO emmylua_code_analysis::profile] analyze workspace main: cost 1.9743589s
[2026-03-27 10:08:27 +08:00 INFO emmylua_code_analysis::profile] update files: cost 0ns
[2026-03-27 10:08:27 +08:00 INFO emmylua_code_analysis::profile] module analyze: cost 2.4218ms
[2026-03-27 10:08:27 +08:00 INFO emmylua_code_analysis::profile] decl analyze: cost 1.2171ms
[2026-03-27 10:08:27 +08:00 INFO emmylua_code_analysis::profile] doc analyze: cost 1.8733ms
[2026-03-27 10:08:27 +08:00 INFO emmylua_code_analysis::profile] flow analyze: cost 164.8µs
[2026-03-27 10:08:27 +08:00 INFO emmylua_code_analysis::profile] lua analyze: cost 905.6µs
[2026-03-27 10:08:27 +08:00 INFO emmylua_code_analysis::profile] resolve analyze: cost 70.2µs
[2026-03-27 10:08:27 +08:00 INFO emmylua_code_analysis::profile] analyze workspace std: cost 4.3118ms
[2026-03-27 10:08:27 +08:00 INFO emmylua_code_analysis::profile] decl analyze: cost 332.9162ms
[2026-03-27 10:08:27 +08:00 INFO emmylua_code_analysis::profile] doc analyze: cost 68.0217ms
[2026-03-27 10:08:27 +08:00 INFO emmylua_code_analysis::profile] flow analyze: cost 93.7479ms
[2026-03-27 10:08:28 +08:00 INFO emmylua_code_analysis::profile] lua analyze: cost 758.1838ms
[2026-03-27 10:08:29 +08:00 INFO emmylua_code_analysis::profile] resolve analyze: cost 602.3406ms
[2026-03-27 10:08:29 +08:00 INFO emmylua_code_analysis::profile] analyze workspace main: cost 1.8554128s
[2026-03-27 10:08:30 +08:00 INFO emmylua_code_analysis::profile] diagnose 1189 files: cost 921.703ms
```

the lastest github release is once:
```log
[2026-03-27 10:17:08 +08:00 INFO emmylua_code_analysis::profile] update files: cost 351.2511ms
[2026-03-27 10:17:08 +08:00 INFO emmylua_code_analysis::profile] module analyze: cost 2.731ms
[2026-03-27 10:17:08 +08:00 INFO emmylua_code_analysis::profile] decl analyze: cost 377.7373ms
[2026-03-27 10:17:08 +08:00 INFO emmylua_code_analysis::profile] doc analyze: cost 73.8883ms
[2026-03-27 10:17:08 +08:00 INFO emmylua_code_analysis::profile] flow analyze: cost 96.0847ms
[2026-03-27 10:17:09 +08:00 INFO emmylua_code_analysis::profile] lua analyze: cost 728.054ms
[2026-03-27 10:17:10 +08:00 INFO emmylua_code_analysis::profile] resolve analyze: cost 529.9525ms
[2026-03-27 10:17:10 +08:00 INFO emmylua_code_analysis::profile] analyze workspace main: cost 1.8059205s
[2026-03-27 10:17:11 +08:00 INFO emmylua_code_analysis::profile] diagnose 1189 files: cost 1.1810153s
```

This should be related to your modifier to the workspace, not this PR. Also, I've noticed memory usage has increased by about 5%, but that's acceptable.

clason:
> the lastest github release is once:

The latest github release also doesn't parse library files under ignored dirs (again); another regression in the latest release that is fixed on `main`.

lewis6991:
> From the startup logs, it looks like the language server parses the workspace twice during startup.

Should be fixed in #995

crates/emmylua_code_analysis/src/semantic/infer/narrow/condition_flow/correlated_flow.rs
```diff
@@ -195,9 +277,10 @@ fn collect_matching_correlated_types(
     discriminant_refs: &[crate::DeclMultiReturnRef],
     target_refs: &[crate::DeclMultiReturnRef],
     narrowed_discriminant_type: &LuaType,
-) -> Result<(Vec<LuaType>, Vec<LuaType>, bool), InferFailReason> {
+) -> Result<(Vec<LuaType>, Vec<LuaType>, Vec<LuaType>, bool, bool), InferFailReason> {
```
CppCXY: Complex return values should use a struct.

crates/emmylua_code_analysis/src/semantic/infer/narrow/condition_flow/correlated_flow.rs
```diff
@@ -278,12 +381,34 @@ fn infer_signature_for_call_ptr<'a>(
...
+                return_type.clone(),
+            );
+            match instantiate_func_generic(db, cache, &func, call_expr) {
+                Ok(instantiated) => instantiated.get_ret().clone(),
+                Err(_) => return_type,
+            }
+        } else {
+            return_type
+        };
+        return vec![crate::LuaSignature::return_type_to_row(
```
CppCXY: clean the full mod path

crates/emmylua_code_analysis/src/semantic/infer/narrow/condition_flow/mod.rs
```diff
@@ -48,6 +55,227 @@ impl InferConditionFlow {
...
+    fn from(result_or_continue: ResultTypeOrContinue) -> Self {
+        match result_or_continue {
+            ResultTypeOrContinue::Continue => ConditionFlowAction::Continue,
+            ResultTypeOrContinue::Result(result_type) => ConditionFlowAction::Result(result_type),
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub(in crate::semantic) enum PendingConditionNarrow {
```
CppCXY, line 85: Rust's enums are powerful, but they're not suitable for representing heavy-weight tree structures unless you add a layer of indirection (e.g., an ID) or use pointers. Note my LuaType — it's limited to a maximum of 16 bytes so we can copy LuaType with low overhead. Here I see you designed two structs that recursively contain each other directly; such a layout can bloat the size of the enum, and it also increases pressure on the stack during destruction. When designing enums, be mindful of the total size of the enum objects. The full-moon library that stylua depends on is a very poor design, causing the AST constructed by full-moon to consume a huge amount of memory.

lewis6991: I've boxed `PendingConditionNarrow` so only references are used. Is this sufficient would you prefer something with ID's or interning?

CppCXY: Depends — if it's expected to form deep tree structures, refer to my FlowTree and FlowNode designs, because enums used for tree structures risk stack-overflow from recursive destruction; otherwise, do as you like.

lewis6991: It seems this has reduced the memory by ~5%.

lewis6991: Ok for now, this is probably ok given how many issues the PR fixes. The design can be refined in follow ups.

CppCXY: Is your object read-only and copyable? If so, consider switching from Box to Arc, because Box owns the object and cloning it will recursively copy deep structures.

lewis6991: It was copying on cache hits and inserts so changed it to Arc.

PR #990, opened 2026-03-20
fix(generic): stabilize higher-order callable overload inference

lewis6991:
Improve overload inference for callable values.

The main rule is: match callable candidates by the actual argument shape before using their return types. That makes higher-order callback inference and direct callable-union calls less order-dependent.

This also clarifies the broad callable cases:
- `unknown` returns stay unknown instead of falling through to a concrete overload.
- bare `function` is treated as an erased callable fallback.
- compatible callable union/intersection members contribute a union of returns.

Regression tests cover `apply`-style callbacks, direct callable unions, and `pcall` return overloads.

### DIFF
--- a/crates/emmylua_code_analysis/src/semantic/generic/instantiate_type/instantiate_func_generic.rs
+++ b/crates/emmylua_code_analysis/src/semantic/generic/instantiate_type/instantiate_func_generic.rs
@@ -7,7 +7,7 @@ use std::{ops::Deref, sync::Arc};
 use crate::semantic::infer::infer_expr_list_types;
 use crate::{
     DocTypeInferContext, FileId, GenericTpl, GenericTplId, LuaFunctionType, LuaGenericType,
-    LuaTypeNode,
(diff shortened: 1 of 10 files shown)


crates/emmylua_code_analysis/src/semantic/generic/instantiate_type/instantiate_func_generic.rs
```diff
@@ -156,60 +157,145 @@ pub fn infer_callable_return_from_remaining_args(
...
-    if tpl_pattern_match_args(
-        &mut callable_context,
-        &callable_param_types,
-        &call_arg_types,
-    )
-    .is_err()
-    {
-        return Ok(Some(fallback_return));
-    }
+    let _ = tpl_pattern_match_args(&mut callable_context, &callable_param_types, call_arg_types);
```
CppCXY: why ignore the err

CppCXY:
please fix conflicts

CppCXY:
please fix conflicts

PR #1013, opened 2026-04-03
fix: support union type in as_doc_function_type

EfveZombie:
Support union types in `as_doc_function_type`, enabling correct `pcall` return type inference when the callee has overloaded members.


EfveZombie:
Closing this — the issue has been properly fixed by #1022, which canonicalizes identical callable unions at the union-building stage. After rebasing this branch onto the latest main (without the fix commit), both test cases pass, confirming that the duplicate callable members from `@field` + implementation are now deduplicated before reaching `as_doc_function_type`, so the union handling added here is no longer needed.

PR #1014, opened 2026-04-03
fix: skip redundant-parameter check when last arg is multi-return call

EfveZombie:
Skip the redundant-parameter diagnostic when the last argument is a multi-return function call (e.g. `table.unpack`), avoiding false positives.


xuhuanzy:
This check is correct and should not be skipped.

PR #1017, opened 2026-04-05
update

### LINKED ISSUE #1007: `constructor` is dependent on load order, and works extremely inconsistently
With the 0.22.0 release, there are suddenly an extreme amount of warnings throughout our codebase which did not exist before; strange warnings such as:

```lua
local MyClass = class("MyClass") -- class uses the constructor attribute properly

return MyClass
```

```lua
--- different file
local myInstance = MyClass() -- WARNING: Cannot call expression of type `EditorDockTree`
```

and:

```lua
---@class MyOtherClass
---@field __instance MyClass
local MyOtherClass = class("MyOtherClass")

MyOtherClass.__instance = MyClass() -- WARNING: Cannot assign `self` to `MyClass`
```

They seem to be related to the order in which things are loaded by EmmyLua; Minimal reproducible examples are very hard for us to create as these issues are very ephemeral.

It seems like the problem occurs when the "constructor defining function" (what uses `@[constructor()]` is "loaded" after an actual call to it.

Nonetheless, attached are two MREs which solely differ in filename; one that always has the problem, and one that never has the problem.

[warning.zip](https://github.com/user-attachments/files/26453388/warning.zip)
[no_warning.zip](https://github.com/user-attachments/files/26453387/no_warning.zip)

Do note that, when testing, only the "main" file should be opened in your IDE; opening one of other two make the issue not occur.

https://github.com/user-attachments/assets/5d0a2606-bc39-4fe4-8083-8bc177b8d2ef

It does seem like these issues have existed since the "constructor" attribute was added, but warnings for "Cannot call expression" did not exist prior.  This system has never worked consistently for us, which is frustrating when our project relies heavily on it. We have noticed that, before this update, the `constructor` attribute would refuse to work most of the time. It has just not triggered warnings until now.

Thank you.

### LINKED ISSUE #1008: Generic class constructors returning raw `self` type
As of the recent EmmyLua update (0.22.0) on Visual Studio Code, constructing a generic class (via the `constructor` attribute) now occassionally returns the `self` type, instead of its actual type.

This seems to happen if the class's generic type is used in its constructor's parameters.

```lua
-- Generic class taking a string in its constructor

---@class ClassA<T>
local ClassA = class("ClassA")

---@param str string
function ClassA:init(str) end


-- Generic class with its generic passed to its constructor

---@class ClassB<T>
local ClassB = class("ClassB")

---@param value T
function ClassB:init(value) end


local a = ClassA("I'm ClassA") -- ClassA<T>
local b = ClassB("I'm ClassB") -- self
```

In this example, `b` should be `ClassB<string>` (or at least some form of `ClassB`).

### LINKED ISSUE #1018: TypeGuards no longer seem to work
```lua
---@class MyClass
---@class MyInheritedClass : MyClass
---@field test fun(): void


---@return TypeGuard<MyInheritedClass>
function typeguard(instance)
    return true
end

---@type MyClass
local instance = {}

if typeguard(instance) then
    instance:test() -- this is MyClass still, this is a warning! Undefined field `test`.

    ---@cast instance MyInheritedClass
    instance:test()
end
```

Tested on the latest update, and also #1017, both have this issue

### LINKED ISSUE #1033: Aliases of generic types aren't working correctly
It seems aliases of generic types don't working correctly at the moment. I'm seeing type-mismatch errors with a message complaining "expected `A`, but got `A` instead":

```lua
--- @namespace generic_alias_type_mismatch

--- @class Foo<T, E>
--- @alias SpecializedFoo<T> Foo<T, string>

--- @generic T
--- @param value T
--- @return Foo<T, any>
function create_foo(value)
  return { value = value } --[[@as Foo<T, any> ]]
end

--- @return SpecializedFoo<string>
function foo()
  return create_foo("bar") --[[@as SpecializedFoo<string> ]]
  --     ^ Annotations specify that return value 1 has a type of
  --     `generic_alias_type_mismatch.SpecializedFoo<string>`, returning value of type
  --     `generic_alias_type_mismatch.SpecializedFoo<string>` here instead.
  --     [return-type-mismatch] ❌
end
```

Reproduced on `0.22.0` and 2f9cb6d5 (HEAD at the time of writing).

### LINKED ISSUE #1049: 更严格的泛型推断
条件泛型目前的推断过于宽泛, `@type`声明泛型类型时允许不带泛型参数, 这给泛型分析带来了不必要的压力.

应学习 ts 的做法将其视为 any 然后报错

```
            ---@alias ExtractFoo<T> T extends { foo: infer F } and F or never

            ---@class Holder<T>
            ---@field value ExtractFoo<T>

            ---@type Holder
            local holder
```

xuhuanzy:
1. #1004 
2. `constructor`特性第四个参数从`return_self`更改为`return_mode`以支持返回值使用泛型
3. 泛型参数现在支持默认值
```lua
---@class A<T = number>
---@class B<T : number = number>
---@class C<T extends number = number> -- `extends` 关键词等效于 `:`, 但更清晰且更容易被大模型识别
```
4. 条件类型完善, 现在条件类型的行为更接近 TS 了


clason:
https://github.com/EmmyLuaLs/emmylua-analyzer-rust/pull/1017/commits/f407e4b31d32d6b1d7667e07ba70fc3c4a66d345 introduced a regression that persists on `HEAD`. The commit message is meaningless to me unfortunately, but I now get a slew of warnings on nvim-treesitter like 
```
-- lua/nvim-treesitter/util.lua [2 warnings]
warning: expected `string? ...` but found `string?`.  [param-type-mismatch]
  -->: lua/nvim-treesitter/util.lua:6:23

  5 │ function M.read_file(filename)
  6 │   local file = assert(io.open(filename, 'r'))
  7 │   local r = file:read('*a')

warning: expected `string? ...` but found `string?`.  [param-type-mismatch]
  -->: lua/nvim-treesitter/util.lua:15:23

  14 │ function M.write_file(filename, content)
  15 │   local file = assert(io.open(filename, 'w'))
  16 │   file:write(content)
  ```

@lewis6991

**Comment by xuhuanzy** [comment]
@clason This is a breaking change. Earlier, I mistakenly treated `extends Type` as a default value constraint, but now it must be written in a syntax similar to TypeScript's `extends Type = Type`. I think this might be the issue.  
Could you provide some reproduction code?

**Comment by clason** [comment]
https://github.com/nvim-treesitter/nvim-treesitter

It's enough to clone this repo and run `VIMRUNTIME=/usr/local/share/nvim/runtime/ emmylua_check .` at the root (assuming you have Neovim installed to `/usr/local`; adjust to taste, e.g., point the `VIMRUNTIME` variable to a local clone of https://github.com/neovim/neovim instead of `/usr/local/share/nvim`).

I am not aware of using type extensions at all.

**Comment by xuhuanzy** [comment]
> [f407e4b](https://github.com/EmmyLuaLs/emmylua-analyzer-rust/commit/f407e4b31d32d6b1d7667e07ba70fc3c4a66d345) introduced a regression that persists on `HEAD`. The commit message is meaningless to me unfortunately, but I now get a slew of warnings on nvim-treesitter like

This issue has nothing to do with generics.
```
--- @type string?
local message

--- @param ... (string?) ...
function test(...) end

test(message)
```

clason:
That may be (and what I assume); nevertheless, I bisected the regression to that commit. Note that the examples I listed deliberately do not involve any user-defined functions; only LuaJIT builtins that emmyluals defines the annotations for.

CppCXY:
> That may be (and what I assume); nevertheless, I bisected the regression to that commit. Note that the examples I listed deliberately do not involve any user-defined functions; only LuaJIT builtins that emmyluals defines the annotations for.

I have time to look into this issue; by the way, why this project was archived?

clason:
> why this project was archived?

Long story I don't want to go much into right now; some people were very vocal in their dislike of the way I handled things, and I didn't want to have to deal with that anymore. The important thing is that it's still available and doing what it was designed to do.

xuhuanzy:
> [f407e4b](https://github.com/EmmyLuaLs/emmylua-analyzer-rust/commit/f407e4b31d32d6b1d7667e07ba70fc3c4a66d345) introduced a regression that persists on `HEAD`. The commit message is meaningless to me unfortunately, but I now get a slew of warnings on nvim-treesitter like

Confirmed it's a generic issue. A long time ago, I used a workaround to treat all such cases as 'any', and this operation has been removed in this commit.


> ```
> --- @type string?
> local message
> 
> --- @param ... (string?) ...
> function test(...) end
> 
> test(message)
> ```

This is an incorrect notation; `---@param ... (string?) ...` should not appear. `...` should only appear after generics. I am considering whether to add a diagnostic or simply remove the `...` keyword, as its semantics are highly unintuitive and cause many processing difficulties.

lewis6991:
`...` should be a keyword. How else would you type annotate varargs?

xuhuanzy:
> `...` should be a keyword. How else would you type annotate varargs?

I am referring to `T...`, and currently, the handling of `T...` is very confusing. Please refer to `fn test_variadic_base()`.

PR #1029, opened 2026-04-09
fix(semantic): make flow narrowing iterative

### LINKED ISSUE #1028: Stuck loading
The current project is stuck loading. I suspect this is because get_field_key is being called incorrectly in some places, since it has O(n²) complexity for arrays.

[maxwellhome.zip](https://github.com/user-attachments/files/26587084/maxwellhome.zip)

lewis6991:
## Summary
- replace recursive flow evaluation with an explicit query stack so deep branch and condition chains stay stack-safe
- cache per-variable flow lookups, flattened branch inputs, and decoded flow metadata to avoid rebuilding narrowing state on repeated queries
- skip redundant `get_field_key()` rescans for value and static table fields during compilation analysis

## Tests
- add regressions for issue #1028, correlated guards, and index-based narrows
- add regressions for the maxwellhome-like large-array case
- add regressions for `pairs()` preserving integer keys on value fields

Fixes #1028

### DIFF
--- a/crates/emmylua_code_analysis/src/compilation/analyzer/lua/stats.rs
+++ b/crates/emmylua_code_analysis/src/compilation/analyzer/lua/stats.rs
@@ -480,82 +480,79 @@ pub fn analyze_local_func_stat(
     Some(())
 }
 
-fn register_expr_key_member(analyzer: &mut LuaAnalyzer, field: &LuaTableField) {
-    // Register expression-key members early so table-decl inference (and pairs)
-    // can see them even when the table itself has no explicit generic type.
-    let Some(field_key) = field.get_field_key() else {
-        return;
-    };
-    let LuaIndexKey::Expr(_) = &field_key else {
-        return;
-    };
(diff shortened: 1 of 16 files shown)


lewis6991:
@CppCXY These changes are quite AI-sloppy but should fix #1028.

If you want something cleaner I can break this up. The main thing is making `get_type_at_flow` iterative so it uses the heap instead of the stack. Even with an iterative approach, the i18n testcase was really slow so I needed to add a bunch of caching to get the perf back to normal.

CppCXY:
I don't think this is a good approach. These data structures are overly complex; we need a more efficient and easier-to-understand design.

lewis6991:
Hmm ok, that might require a larger rewrite to remove the recursion.

lewis6991:
Have a look at this now. I've rewritten `get_type_at_flow.rs` to use `FlowTypeEngine`. I think this is much easieer to understand now.

xuhuanzy:
@CppCXY seems to be planning to rewrite the entire analyzer on the `salsa` branch. I suggest asking for his input before moving on to the next step.

PR #1031, opened 2026-04-10
fix: support bracket index in @cast expressions, to enable type narrowing for array element access

EfveZombie:
## Problem

`@cast` expressions with bracket index access (e.g., `---@cast addresses[1] -nil`) were not working correctly. Two issues existed:

### 1. Lexer/Parser did not support `[]` in `@cast` expressions
The doc lexer's `lex_cast_expr` state did not recognize `[`, `]`, or integer literals, causing `---@cast addresses[1] -nil` to be incorrectly parsed. This corrupted the type inference for `addresses`, leading to a spurious `Undefined field [1]` warning.

### 2. Parsed `IndexExpr` had wrong AST structure
After fixing the parser, the `TkInt` token inside `[1]` was placed as a bare token in the `IndexExpr` node. However, `LuaIndexExpr::get_index_key()` expects a child **Node** (like `LuaLiteralExpr`), not a bare token. This caused:
- `get_index_key()` → `None`
- `get_access_path()` → `None`
- `get_index_expr_var_ref_id()` → `None`
- The cast's type narrowing silently failed to take effect
- `addresses[1]` remained `string|nil` instead of being narrowed to `string`, causing a spurious `param-type-mismatch` warning

## Fix

1. **Lexer** (`lua_doc_lexer.rs`): Added support for `[`, `]`, and digit literals in `lex_cast_expr`
2. **Parser** (`tag.rs`): In `parse_cast_expr`, wrap `TkInt`/`TkString` tokens inside `[]` in a `LiteralExpr` node, making the doc `IndexExpr` structure consistent with Lua code `IndexExpr`

## Tests

- Added `test_array_index_with_cast` in `undefined_field_test.rs`
- Added `test_cast_bracket_index_narrows_type` in `param_type_check_test.rs` to verify the narrowing actually works
- All 326 diagnostic tests pass

### DIFF
--- a/crates/emmylua_parser/src/lexer/lua_doc_lexer.rs
+++ b/crates/emmylua_parser/src/lexer/lua_doc_lexer.rs
@@ -599,10 +599,23 @@ impl LuaDocLexer<'_> {
                 reader.bump();
                 LuaTokenKind::TkDot
             }
+            '[' => {
+                reader.bump();
+                LuaTokenKind::TkLeftBracket
+            }
+            ']' => {
+                reader.bump();
+                LuaTokenKind::TkRightBracket
+            }
             ch if is_name_start(ch) => {
                 reader.bump();
                 reader.eat_while(is_name_continue);
                 LuaTokenKind::TkName
+            }
+            '0'..='9' => {
+                reader.bump();
+                reader.eat_while(|c| c.is_ascii_digit());
+                LuaTokenKind::TkInt
             }
             _ => self.lex_normal(),
         }
--- a/crates/emmylua_code_analysis/src/diagnostic/test/param_type_check_test.rs
+++ b/crates/emmylua_code_analysis/src/diagnostic/test/param_type_check_test.rs
@@ -1555,4 +1555,36 @@ mod test {
         "#,
         ));
     }
+
+    #[test]
+    fn test_cast_bracket_index_narrows_type() {
+        let mut ws = VirtualWorkspace::new();
+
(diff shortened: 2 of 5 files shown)


xuhuanzy:
This shouldn’t be handled this way; I’ve decided to implement it myself.

xuhuanzy:
Since cppcxy is rewriting the entire analyzer, `cast` will not be rewritten for the time being.

crates/emmylua_parser/src/lexer/lua_doc_lexer.rs
```diff
@@ -599,11 +599,24 @@ impl LuaDocLexer<'_> {
...
+            ']' => {
+                reader.bump();
+                LuaTokenKind::TkRightBracket
+            }
             ch if is_name_start(ch) => {
                 reader.bump();
                 reader.eat_while(is_name_continue);
                 LuaTokenKind::TkName
             }
+            ch if ch.is_ascii_digit() => {
```
CppCXY, line 615: wrong if '0z'

EfveZombie: Good catch! Fixed by using explicit char range `'0'..='9'` instead of `is_ascii_digit()`, so only pure digit characters enter the integer branch. Also added a parser-level AST test (`test_cast_bracket_index`) covering `[1]`, `["key"]`, `[name]`, and mixed `a.b[1].c` to verify correct node wrapping.

The branch has been rebased onto latest `main` and split into two commits: tests first, then fix.

PR #1037, opened 2026-04-16
fix(flow): reuse ordered return analysis for loop exits

lewis6991:
Replace the old missing-return special cases with the same ordered block
walker used for inferred returns. This avoids false MissingReturn
diagnostics for functions that return from truthy while/repeat loops,
keeps break and repeat-until flow consistent across inference and
diagnostics, and adds regressions for closure returns, diagnostics, and
module exports.

### DIFF
--- a/crates/emmylua_code_analysis/src/compilation/analyzer/lua/closure.rs
+++ b/crates/emmylua_code_analysis/src/compilation/analyzer/lua/closure.rs
@@ -14,7 +14,7 @@ use crate::{
     infer_expr,
 };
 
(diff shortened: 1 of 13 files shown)


clason:
regression in [nvim-treesitter's `indent.lua`](https://github.com/nvim-treesitter/nvim-treesitter/blob/main/lua/nvim-treesitter/indent.lua): spurious
> Annotations specify that a return value is required here.

at line 344.

lewis6991:
#1039

PR #1044, opened 2026-04-20
fix(semantic): avoid recursive inference while preserving flow precision

lewis6991:
Flow sensitive narrowing previously reached back into top level
expression and call inference from inside the flow engine. Condition
handling, correlated call lookup, initializer recovery, and call probing
could ask questions that were already in flight, which opened a
recursion path and forced conservative bailouts around the hot spots.

Fix that by splitting normal and no flow expression and call caches.
Ready remains the in flight sentinel, while Declined is scoped to no
flow caches and records that a static probe already failed and must not
recurse. This makes failed no flow lookups sticky and lets static
callers stop cleanly instead of re entering full inference.

Recover the lost precision by routing flow sensitive leaf evaluation
back through the existing flow scheduler. NameExpr and IndexExpr now
resolve through get_type_at_flow first, then fall back to no flow only
when the scheduler cannot answer. Reuse that bridge from binary
condition narrowing, call condition narrowing, correlated signature
lookup, and declaration initializer recovery so branch local facts still
apply to equality checks, aliased pcall and rawget guards, and local y =
x inside narrowed branches.

Keep the remaining ambiguity policy local to overload resolution. When a
no flow argument declines, treat it as compatible but non
disambiguating. Only choose an overload when the surviving set is unique
or every remaining candidate returns the same type, and decline mixed
return ambiguity instead of guessing.

### DIFF
--- a/crates/emmylua_code_analysis/src/compilation/test/flow.rs
+++ b/crates/emmylua_code_analysis/src/compilation/test/flow.rs
@@ -193,6 +193,38 @@ mod test {
     }
 
     #[test]
+    fn test_stacked_local_call_alias_type_guards_build_semantic_model() {
+        let mut ws = VirtualWorkspace::new();
+        let repeated_guards = "if not pred(value) then return end\n".repeat(STACKED_TYPE_GUARDS);
+        let block = format!(
+            r#"
+        ---@param v any
+        ---@return TypeGuard<string>
+        local function is_string(v)
+            return true
+        end
+
+        local pred = is_string
+        local value ---@type string|integer|boolean
+
+        {repeated_guards}
+        after_guard = value
+        "#,
+        );
+
+        let file_id = ws.def(&block);
+
+        assert!(
+            ws.analysis
+                .compilation
+                .get_semantic_model(file_id)
+                .is_some(),
+            "expected semantic model for stacked local call alias type guard repro"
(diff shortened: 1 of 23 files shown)


CppCXY:
SCC is also a feature that my salsa branch plans to introduce. Since you have chosen to introduce SCC, please prove that its parsing performance will be better. Also, when referencing types, please use short names instead of crate::<path>.

lewis6991:
sorry, I wasn't supposed to push the last change.

lewis6991:
> SCC is also a feature that my salsa branch plans to introduce. Since you have chosen to introduce SCC, please prove that its parsing performance will be better. Also, when referencing types, please use short names instead of crate::.

What are the plans and status of the salsa branch? Is it something you plan to merge in soon?

CppCXY:
> > SCC is also a feature that my salsa branch plans to introduce. Since you have chosen to introduce SCC, please prove that its parsing performance will be better. Also, when referencing types, please use short names instead of crate::.
> 
> What are the plans and status of the salsa branch? Is it something you plan to merge in soon?

No, that won't be the case. I plan to remove the entire db_index when introducing salsa, and that won't be completed until much later. So feel free to make any changes.

lewis6991:
Replaced by #1057.

PR #1051, opened 2026-04-27
fix(semantic): stop following circular type declarations

lewis6991:
When aliases or class parents point back to themselves, stop using that
loop for normal analysis instead of trying to expand it forever.

Pure alias loops now become any, and circular class parents are still
visible to the existing diagnostic.


crates/emmylua_code_analysis/src/compilation/analyzer/doc/type_def_tags.rs
```diff
@@ -173,6 +175,40 @@ pub fn analyze_alias(analyzer: &mut DocAnalyzer, tag: LuaDocTagAlias) -> Option<
     Some(())
 }
 
+fn alias_origin_reaches(db: &crate::DbIndex, origin: &LuaType, target_id: &LuaTypeDeclId) -> bool {
```
CppCXY: remove crate::

PR #1054, opened 2026-04-29
test: make diagnostic helper name match behavior

lewis6991:
Rename check_code_for to has_no_diagnostic so assertions describe the
existing behavior directly.

Assisted-by: Codex

### DIFF
--- a/crates/emmylua_code_analysis/src/compilation/test/and_or_test.rs
+++ b/crates/emmylua_code_analysis/src/compilation/test/and_or_test.rs
@@ -6,7 +6,7 @@ mod test {
     fn test_issue_221() {
         let mut ws = VirtualWorkspace::new();
 
-        assert!(ws.check_code_for(
+        assert!(ws.has_no_diagnostic(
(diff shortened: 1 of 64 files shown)


CppCXY:
If the logic is to be completely reversed, that usually means the current name is not appropriate. I would suggest renaming it rather than reversing the judgment logic.

lewis6991:
A better name with the old logic would be `check_no_code_for`, but that just embeds the `!` into the name. I do think keeping the name and reversing the result is better here. `check_code_for` means "run this code and check it raises this diagnostic".

CppCXY:
> A better name with the old logic would be , but that just embeds the into the name. I do think keeping the name and reversing the result is better here. means "run this code and check it raises this diagnostic".`check_no_code_for``!``check_code_for`

The problem with reversing the logic is that it would result in unmerged code, which would naturally be reversed in terms of logic precisely because of this PR, and cannot be easily fixed by simple text replacement.

PR #1086, opened 2026-05-14
fix: handle semantic flow edge cases

lewis6991:
Problem:
- Several semantic-flow edge cases produced false diagnostics or imprecise types.
- The fixes cover union array keys, empty generic varargs, non-terminating return paths, guarded pcall callbacks, and call-condition replay.

Solution:
- Keep the fixes split into focused commits with targeted regressions.
- Tighten flow replay and branch narrowing where the existing model already had the right seam.

Testing:
- cargo test -p emmylua_code_analysis


lewis6991:
@CppCXY friendly ping. This is just a collection of small fixes (split into commits) I found when running through https://github.com/lewis6991/gitsigns.nvim

PR #1089, opened 2026-05-18
fix: make code actions follow formatter tag spacing

lewis6991:
## Problem

`codeAction.insertSpace` is currently a boolean defaulting to false, so generated diagnostic comments cannot follow the formatter configured EmmyLua tag spacing.

## Solution

Make `codeAction.insertSpace` optional. When omitted, generated diagnostic comments resolve the formatter config for the current file and use `emmy_doc.space_between_tag_columns`. Explicit `codeAction.insertSpace` still overrides the formatter setting.

Update the generated schema and config docs.

### DIFF
--- a/crates/emmylua_code_analysis/resources/schema.json
+++ b/crates/emmylua_code_analysis/resources/schema.json
@@ -12,7 +12,7 @@
     "codeAction": {
       "$ref": "#/$defs/EmmyrcCodeAction",
       "default": {
-        "insertSpace": false
+        "insertSpace": null
       }
     },
     "codeLens": {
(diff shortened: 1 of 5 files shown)


CppCXY:
What is the difference between this option and space_between_tag_columns? Maybe my name isn't very appropriate, but it should be for this function.

Edit: I reviewed the code again. If this configuration option is placed here, it seems it cannot be used with the external formatter, so is it really appropriate?

lewis6991:
You're right, sorry. I didn't look closely enough at what already existed.

`format.insertSpace` was just duplicating `emmy_doc.space_between_tag_columns`, and your external formatter point is valid too: that option would not have affected external formatter output.

I've revised the PR to remove the new formatter option. The remaining change only makes `codeAction.insertSpace` optional; when it is omitted, code actions now fall back to the existing resolved formatter config for the file, `emmy_doc.space_between_tag_columns`.

PR #1098, opened 2026-06-03
fix: separate unreachable conditions from never types

lewis6991:
## Problem

Flow queries used `never` for two different ideas:

- a value whose real type is `never`
- a condition edge that cannot be reached

Those are not equivalent. A query for a symbol inside an impossible branch still needs to be able to walk past the impossible condition and find the symbol's declared or assigned type.

For example, this branch is unreachable, but `value` itself is still a `string`:

```lua
---@return false
local function always_false()
    return false
end

---@type string
local value = "ok"

if always_false() then
    inner = value -- should be string, not never
end
```

The same issue could show up as bogus diagnostics for unrelated call targets inside impossible branches:

```lua
---@return true
local function always_true()
    return true
end

local function b() end

if always_true() then
else
    b() -- should not become call-non-callable
end
```

At the same time, merge queries must still treat impossible predecessors as contributing nothing. In this case, the assignment in the impossible branch must not leak into the joined type:

```lua
---@return false
local function always_false()
    return false
end

local value = "before"

if always_false() then
    value = 1
end

after = value -- should be string, not string|1
```

## Solution

Separate unreachable condition edges from ordinary `never` values in flow-query results:

```rust
enum FlowQueryResult {
    Type(LuaType),
    Unreachable,
}
```

Then interpret `Unreachable` based on the kind of query being answered:

```rust
match flow_mode {
    FlowMode::Normal => continue_to_antecedent(),
    FlowMode::IgnoreConditions => continue_to_antecedent(),
    FlowMode::MergeBranch => contribute_never_to_merge(),
}
```

That gives the two cases different behavior:

- Point queries continue through impossible condition edges so unrelated symbols keep their real type.
- Merge-branch queries turn impossible predecessors into `never`, so unreachable assignments, casts, annotations, and fallback RHS values do not contribute to the joined type.

The change also preserves `MergeBranch` mode through the paths that can otherwise reintroduce unreachable branch facts:

```lua
local value = "before"

if always_false() then
    value = 1                  -- assignment ignored at merge
    ---@cast value integer      -- cast ignored at merge
    ---@type integer
    value = 1                  -- annotated assignment ignored at merge
end

after = value -- remains string
```

Reachable real `never` values are still distinct from unreachable branches. This still works because `never` is not treated as an unreachable edge by itself:

```lua
---@class NeverBox
---@field value never

---@return NeverBox
local function make_box() end

local value = make_box().value
local cond ---@type boolean

if cond then
    value = 1
end

after = value -- should be 1
```

### DIFF
--- a/crates/emmylua_code_analysis/src/compilation/analyzer/flow/bind_analyze/stats.rs
+++ b/crates/emmylua_code_analysis/src/compilation/analyzer/flow/bind_analyze/stats.rs
@@ -199,7 +199,9 @@ pub fn bind_call_expr_stat(
         if let Some(ast) = LuaAst::cast(call_expr.syntax().clone()) {
             bind_each_child(binder, ast, current);
         }
-        current
+        let flow_id = binder.create_node(FlowNodeKind::CallExprStat(call_expr_stat.to_ptr()));
+        binder.add_antecedent(flow_id, current);
+        flow_id
     }
 }
 
--- a/crates/emmylua_code_analysis/src/compilation/test/flow.rs
+++ b/crates/emmylua_code_analysis/src/compilation/test/flow.rs
@@ -308,6 +308,180 @@ mod test {
     }
 
     #[test]
+    fn test_plain_call_condition_keeps_inner_call_prefix_type() {
+        let mut ws = VirtualWorkspace::new();
+        let code = r#"
+        local function a() end
+        local function b() end
+
+        b()
+        if a() then
+            b()
+            inner = b
+        end
+        "#;
+        ws.def(code);
+
+        let ty = ws.expr_ty("inner");
+        assert!(ty.is_function());
+
+        let mut diag_ws = VirtualWorkspace::new();
+        assert!(diag_ws.has_no_diagnostic(DiagnosticCode::CallNonCallable, code));
+    }
+
+    #[test]
+    fn test_false_call_condition_keeps_inner_unrelated_type() {
+        let mut ws = VirtualWorkspace::new();
+        ws.def(
+            r#"
+        ---@return false
+        local function always_false()
+            return false
+        end
+
+        ---@type string
+        local value = "ok"
+        if always_false() then
+            inner = value
+        end
+        "#,
+        );
+
+        assert_eq!(ws.expr_ty("inner"), ws.ty("string"));
+    }
+
+    #[test]
+    fn test_true_call_condition_keeps_else_call_prefix_type() {
+        let mut ws = VirtualWorkspace::new();
+        let code = r#"
+        ---@return true
+        local function always_true()
+            return true
+        end
+
+        local function b() end
+        if always_true() then
+        else
+            b()
+        end
+        "#;
+
+        assert!(ws.has_no_diagnostic(DiagnosticCode::CallNonCallable, code));
+    }
+
+    #[test]
+    fn test_false_call_condition_assignment_does_not_contribute_to_merge() {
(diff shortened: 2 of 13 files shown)


lewis6991:
@CppCXY friendly ping. This change overlaps with #1113.

xuhuanzy:
> This change overlaps with #1113.

Since this is a massive change, I will make adjustments after using it for a while, so this PR can be merged first.

xuhuanzy:
Now `val` is `string`, but I think it should be `never`. What do you think?
```
---@type number
local val

if type(val) == "string" then
    print(val) -- is `never`
end
```

lewis6991:
I think it should be `string`, but the whole block is marked as unreachable. If you mark as `never`, then it can create some noisy diagnostics, whereas the only diagnostic that matters is the block being unreachable.

xuhuanzy:
I think it's best to use `never`; that's how it is in TypeScript. I can handle the diagnostics.

lewis6991:
I really don't think so. In the code above the `---@type number` is just a type hint, and could be wrong. If `type(val) == "string"` succeeds, then `val` has to be `string`. It's a fact we know for certain. We don't know `val` can be `number`, the annotation just describes programmer intent.

TypeScript is not a strong precedent here because TS treats the declared type as the authoritative static domain. EmmyLua is modeling dynamic Lua with annotations layered on top. In that world, runtime guards should be allowed to describe what would be true if execution reached the block.

xuhuanzy:
What if, instead of the `type` function, we're using other custom type guards? I know what you mean, but I believe we should trust the annotations. For now, though, sticking with the current situation seems like the right call.

lewis6991:
Yeah, I get that `never` avoids choosing between the two annotations, but I think that is the wrong tradeoff. It turns a local conflict into a poisoned value type.

I think we should choose the most local fact for point queries. The declared type is the baseline, but a guard/cast/assignment is the fact being asserted at that exact program point, so it should win inside that branch.

Reachability should stay separate. If the local fact conflicts with the baseline, we can mark/report the branch as unreachable or inconsistent, but the value inside the branch should still use the local fact. That avoids cascading `never` diagnostics while preserving the useful signal.

xuhuanzy:
Actually, this is a similar issue I encountered while writing the wiki.  
Now an error has occurred, but under `@cast`, this error should not happen. However, what about during type guarding? If it is still allowed, does this diagnostic even need to exist?
```lua
---@type string
local val

---@cast val + boolean
val = true --  assign-type-mismatch
```

xuhuanzy:
To address this issue, we might need to change `no_flow_mode` to an enum to allow flow execution in certain cases. However, this would increase complexity. Do you have any suggestions?

lewis6991:
In this case I would expect no diagnostic. The `@cast` depends on the previous `@type` fact and says "add boolean as an allowed type of `val`", so the effective local type becomes `string|boolean`.

So I think the same rule still applies: the most local fact wins for point queries, but that local fact can be derived from earlier facts. Here the local fact is not just `boolean`; it is the previous type plus `boolean`.

That said, this PR does not address this and reports `assign-type-mismatch`, so I think this is a separate diagnostics issue from unreachable-vs-`never`.

lewis6991:
> To address this issue, we might need to change `no_flow_mode` to an enum to allow flow execution in certain cases. However, this would increase complexity. Do you have any suggestions?

Fixed in https://github.com/EmmyLuaLs/emmylua-analyzer-rust/pull/1098/changes/f399a1ff1987ab48983656f1461d16a1d5cb4a3d

PR #1107, opened 2026-06-11
fix: add Union branch to check_generic_type_compact to prevent param-type-mismatch false positives

### LINKED ISSUE #1106: `param-type-mismatch` / `assign-type-mismatch` false positive when a method returning a generic type has both `@field` declaration and function implementation
### Summary

When a method that returns a generic type (e.g., `Box<Item>`) has **both** an `@field fun():Generic<X>` declaration and a `function` implementation, the return value is internally wrapped in a Union type. The generic constraint checker `check_generic_type_compact` lacks a `LuaType::Union` branch, causing it to reject the Union-wrapped type as incompatible — even though the underlying types are identical. This produces `param-type-mismatch` or `assign-type-mismatch` false positives.

### Minimal Reproduction

```lua
---@class Item
---@field id integer

---@class Box<T>
---@field value T

---@class Service
---@field execute fun(self:Service, box:Box<Item>)
---@field fetch fun(self:Service):Box<Item>
local Service = {}

function Service:execute(_box)
end

function Service:fetch()
    return {} --[[@as Box<Item>]]
end

function Service:run()
    local b = self:fetch()
    self:execute(b)  -- ❌ False positive: expected `Box<Item>` but found `Box<Item>` [param-type-mismatch]
end
```

**Key observations:**
- The bug only triggers when the **value-producing method** (`fetch`) has both `@field` declaration and `function` implementation. The consuming method (`execute`) is irrelevant.
- The bug occurs with both parameter passing (`param-type-mismatch`) and assignment (`assign-type-mismatch`).
- If `fetch` has only `@field` (no implementation) or only `function` + `@return` (no `@field`), the bug does not occur.
- The bug does not occur when the return type is non-generic (e.g., `string`).

### Expected Behavior

`self:execute(b)` should produce no warnings because `b`'s type `Box<Item>` matches the parameter declaration exactly.

### Actual Behavior

```text
warning: expected `Box<Item>` but found `Box<Item>`. [param-type-mismatch]
```

The error message shows identical types, but internally the actual type is `Box<Item> | Box<Item>` (a Union wrapping `Box<Item>`), which the generic checker cannot handle.

### Reproduction Steps

1. Create a minimal reproduction file `bug1_repro.lua`:

   ```lua
   ---@class Item
   ---@field id integer
   ---@class Box<T>
   ---@field value T
   ---@class Service
   ---@field execute fun(self:Service, box:Box<Item>)
   ---@field fetch fun(self:Service):Box<Item>
   local Service = {}
   function Service:execute(_box) end
   function Service:fetch() return {} --[[@as Box<Item>]] end
   function Service:run()
       local b = self:fetch()
       self:execute(b)
   end
   ```

2. Run the check:

   ```bash
   emmylua_check bug1_repro.lua
   ```

3. Observe the false positive:

   ```text
   ---  [1 warning]
   warning: expected `Box<Item>` but found `Box<Item>`. [param-type-mismatch]
     --> :17:18

    16 |     local b = self:fetch()
    17 |     self:execute(b)

   Summary
     1 warning

   Check completed with warnings
   Check finished
   ```

### Root Cause

The issue has two layers:

**Layer 1: Union creation.** When `fetch` has both `@field fun():Box<Item>` (marked as `FileFieldDecl`) and `function Service:fetch()` (marked as `FileMethodDecl`), `resolve_member_type` in `lua_member_item.rs` treats both as `is_file_decl() == true` and merges them via `union_all`, producing `fun():Box<Item> | fun():Box<Item>`. Calling `self:fetch()` then returns a Union-wrapped `Box<Item>`.

**Layer 2: Missing Union branch.** In `check_generic_type_compact` (`generic_type.rs`), the `match` statement handles `GenericType`, `TableConst`, `Ref`/`Def`, but has **no `LuaType::Union` branch**. The Union-wrapped type falls through to `_ => Err(TypeCheckFailReason::TypeNotMatch)`, producing the false positive.

### Environment

- **emmylua_check version**: based on commit `d7d5dcbf`
- **OS**: Linux

ShenzhenGopher:
## Summary

Fixes #1106 

When a method returning a generic type has both `@field` declaration and `function` implementation, the return value is internally wrapped in a Union type. `check_generic_type_compact` lacks a `LuaType::Union` branch, causing false `param-type-mismatch` / `assign-type-mismatch` warnings even when types are identical.

## Root Cause

In `check_generic_type_compact` (`crates/emmylua_code_analysis/src/semantic/type_check/generic_type.rs`, line 99), the `match` statement handles `GenericType`, `TableConst`, `Ref`/`Def`, but has no `LuaType::Union` branch. When `compact_type` is a Union (produced by merging `@field` and `function` sources in `resolve_member_type`), it falls through to `_ => Err(TypeCheckFailReason::TypeNotMatch)`.

## Changes

| File | Change |
|------|--------|
| `crates/emmylua_code_analysis/src/semantic/type_check/generic_type.rs` | Add `LuaType::Union` branch in `check_generic_type_compact` that iterates all union members and checks each against the source generic type (+13 lines) |

## Code Change

Insert before the `_ => Err(TypeCheckFailReason::TypeNotMatch)` arm in `check_generic_type_compact`:

```rust
        LuaType::Union(union_type) => {
            // When compact_type is a Union, every member of the union must be
            // assignable to the source generic type for the assignment to be safe.
            for member_type in union_type.into_vec() {
                check_generic_type_compact(
                    context,
                    source_generic,
                    &member_type,
                    check_guard.next_level()?,
                )?;
            }
            Ok(())
        }
```

## Rationale

A Union type `A | B` assigned to a generic parameter `T` requires **all** members to satisfy `T`'s constraints for the assignment to be safe. This is standard covariant semantics — the Union is a "wider" type, so each member must individually be valid.

## Example

```lua
---@class Item
---@field id integer
---@class Box<T>
---@field value T
---@class Service
---@field execute fun(self:Service, box:Box<Item>)
---@field fetch fun(self:Service):Box<Item>
local Service = {}
function Service:execute(_box) end
function Service:fetch() return {} --[[@as Box<Item>]] end
function Service:run()
    local b = self:fetch()
    self:execute(b)
end
```

## Verification

Before fix (commit `d7d5dcbf`):

```text
---  [1 warning]
warning: expected `Box<Item>` but found `Box<Item>`. [param-type-mismatch]
  --> :17:18

 16 |     local b = self:fetch()
 17 |     self:execute(b)

Summary
  1 warning

Check completed with warnings
Check finished
```

After fix:

```text
No issues found
Check finished
```

### DIFF
--- a/crates/emmylua_code_analysis/src/semantic/type_check/generic_type.rs
+++ b/crates/emmylua_code_analysis/src/semantic/type_check/generic_type.rs
@@ -99,6 +99,19 @@ pub fn check_generic_type_compact(
                 check_guard.next_level()?,
             )
         }
+        LuaType::Union(union_type) => {
+            // When compact_type is a Union, every member of the union must be
+            // assignable to the source generic type for the assignment to be safe.
+            for member_type in union_type.into_vec() {
+                check_generic_type_compact(
+                    context,
+                    source_generic,
+                    &member_type,
+                    check_guard.next_level()?,
+                )?;
+            }
+            Ok(())
+        }
         _ => Err(TypeCheckFailReason::TypeNotMatch),
     }
 }


xuhuanzy:
All changes must include tests, and since the generic version is currently unstable, approval is not being considered for now.

Issue #877, opened 2025-12-08
`never` doesn't do what it's supposed to do


idanarye:
Using version 0.18 downloaded from the GH release:
```
$ ./emmylua_check --version
emmylua_check 0.18.0
```

I wrote this file:

```lua
---@return integer?
local function create_num()
    return 1
end

---@param _num integer
local function accept_num(_num)
end

---@return never
local function panic()
    error("panic!")
end

local num1 = create_num()
if num1 == nil then
    panic()
end
accept_num(num1)

local num2 = create_num()
if num2 == nil then
    error()
end
accept_num(num2)
```

And ran:
```
$ ./emmylua_check .
--- main.lua [1 warning]
warning: expected `integer` but found `integer?`.  [param-type-mismatch]
  -->: main.lua:19:12

  18 │ end
  19 │ accept_num(num1)
  20 │ 



Summary
  1 warning

Check completed with warnings
Check finished
```

`num2` worked as expected - using the builtin `error` when its `nil` convinced the type checker that it cannot be `nil`. But why does it think `num1` is still an `integer?`? I've annotated my `panic()` as `---@return never`, which - in any language that has a never type - means it can never return (either by looping forever, by throwing an error, or by crashing the program itself via whatever means), and therefore the type checker should consider it equivalent to calling `error()` and conclude the `num1` is `integer` and not `integer?` after that check. But it doesn't.

CppCXY:
you can config like:
```json
{
    "runtime": {
        "special": {
            "panic" : "error"
        },
        "version": "LuaLatest"
    }
}

```

In the current control-flow analysis we assume our type analysis is not yet complete, so we can't make use of that information.

idanarye:
What's the purpose of the `never` type, then?

xuhuanzy:
`never` is currently equivalent to `nil`

xuhuanzy:
I think it's now possible to support this issue to a limited extent.

xuhuanzy:
Added a rough implementation without performing detailed control flow analysis.

PR #1113, opened 2026-06-14
update

### LINKED ISSUE #560: 枚举值运算问题
在 ts 中, 枚举值运算的结果不是实际值, 但我们目前的结果是实际值, 所以难以处理
```lua
            ---@enum SubscriberFlags
            local SubscriberFlags = {
                Tracking = 1 << 0,
            }
            ---@class Subscriber
            ---@field flags SubscriberFlags
            
            ---@type Subscriber
            local subscriber
            
            subscriber.flags = subscriber.flags & ~SubscriberFlags.Tracking
            
            subscriber.flags = 9  -- 应该警告
```

### LINKED ISSUE #616: How do you capture a generic in a constructor?
I'm trying to make a state machine class, where it should capture the state using the arguments.

```lua
---@class StateMachine<T : State> : class
local StateMachine = class("StateMachine")

---@generic T : State
---@param state_type T
function StateMachine:init(state_type)
    -- Use it in here
end
```

Unfortunately, just calling `local machine = StateMachine(TestState)` does not make a `StateMachine<TestState>`, only a `StateMachine`. Is there a way I can annotate this properly so it works? Ideally I wouldn't need to `---@type StateMachine<TestState>` above the line...

For reference, our config:

```json
    "classDefaultCall": {
      "forceNonColon": true,
      "forceReturnSelf": true,
      "functionName": "init"
    },
```

### LINKED ISSUE #626: FlowNodeKind::BranchLabel 应该要有条件地联合
```rust
    #[test]
    fn test_table_narrow() {
        let mut ws = VirtualWorkspace::new();
        ws.def(
            r#"
                ---@type table<int, table<int,table>>
                local archiveCache = {}

                local playerCache = archiveCache[0]
                if not playerCache then
                    playerCache = {}
                end

                A = playerCache -- 此时是 table | table<integer,table>, 但他应该为 table<integer,table>
            "#,
        );
        let a = ws.expr_ty("A");
        assert_eq!(ws.humanize_type(a), "table<integer,table>");
    }
```

### LINKED ISSUE #632: classDefaultCall引用计数不正确
现在用的地方

Image可以正确标识new，也可以从new跳到ctor那边去，但是ctor方法上的引用为0

Image，这个感觉怎么处理一下比较舒服； 然后就是有多个ctor重载的时候，参数提示没有重载的选项

Image

Image即使写了第一个参数，也不会出现

Image

### LINKED ISSUE #711: Add option to not show all overloads in hint
it would be nice if there was a configuration option to consolidate overload function definitions in hints. Especially when they have the same number of arguments and underlying argument type

Image

selecting any of these hitnts results in the same thing being appended to my lua file.

Image

from there i still want all my overloads for argument suggestion (i like this behavior)

Image

global functions already behave this way

Image

here is the def for the example above

Image

basically im adding exact cases to a generic where i dont want all classes that inherent a base as suggestions

### LINKED ISSUE #860: alias 包含nil时的bug
---@alias PersonAge
---| integer
---| nil

---@class Person
---@field name string
---@field age PersonAge

---@type Person
local person = {
    name = "123",
}

### LINKED ISSUE #903: lhs must be non-nil when assert in rhs
```lua
---@type integer[][]
local res
res[1][1] = assert(res[1])[2]
```

```
1. res[1] may be nil [need-check-nil]
```

### LINKED ISSUE #910: type is lost when using param in the return statement of lambda with generic return
It seems like the type for a parameter is lost if it's used in the return statement of a lambda function with a generic parameter, and return (i.e. `fun<T, U>(v: T): U`):

```lua
--- @generic T
--- @param ... T...
--- @return T...
local function identity(...) return ... end

--- @generic T, U
--- @param list T[]
--- @param fn fun(item: T): U
--- @return U[]
local function map(list, fn) return nil --[[@as U[] ]] end

local list_1 --- @type string[]

local _mapped_1 =
  --  ^ type: `int[]` ✅
  map(list_1, function(item)
    local _a = identity(item)
    --    ^ type: `string` ✅
    --    The type is correct while the param doesn't appear in the return
    --    statement.

    return 42
  end)

local _mapped_2 =
  --  ^ type: `any[]` ❌ should be `string[]`
  map(list_1, function(item)
    local _a = identity(item)
    --    ^ type: `any` ❌
    --    The type is lost if the param is used in the return statement.

    return item
  end)

local _mapped_3 =
  --  ^ type: `string[]` ✅
  map(list_1, function(item)
    local _a = identity(item)
    --    ^ type: `string` ✅
    --    The type is correct if the return is explicitly typed.

    return item --[[@as string ]]
  end)
```

### LINKED ISSUE #912: no narrow after `if not (a and b)`
```lua
local a, b = ("abc"):find("..")
-- this work
-- if not a or not b then
-- this don't
if not (a and b) then
	return
end

-- should be unnecessary assert
assert(a)
```

### LINKED ISSUE #923: goto label give wrong reference
`/bin/nvim --clean -u a.lua a.lua`
```lua
local M = {}
M.cont = function() end
vim.lsp.config("ls", {
	cmd = { "emmylua_ls" },
	-- cmd = { 'lua-language-server' },
	filetypes = { "lua" },
	root_markers = {
		".luarc.json",
		".emmyrc.json",
		".luacheckrc",
		".git",
	},
	settings = { Lua = { workspace = { library = { "$VIMRUNTIME" } } } },
	workspace_required = false,
})
vim.lsp.enable("ls")

if false then
	while true do
		goto cont
		::cont::
	end
end
```

goto reference on `::cont::`, shouldn't give reference on function

### LINKED ISSUE #956: Missing class documentation
Migrating from sumnekos LuaLS, I noticed a change in the way class documentation strings are handled.

For my meta libraries, I used to write code in a similar pattern like this:

```lua
--- This is the MyModule documentation.
--- It should appear when hovering over MyModule.
--- @class MyModule
MyModule = {}


--- This is the SubModule documentation.
--- It should appear when hovering over SubModule.
--- @class MyModule.SubModule
MyModule.SubModule = {}

--- Creates a new SubModule instance
--- @return MyModule.SubModule
function MyModule.SubModule.create()
end

return MyModule
```

Using the Emmy LS, I now noticed that my documentation strings are not shown to the user when hovering over class usages.

To better showcase my issue, here is how the Emmy LS displays the hover message vs Sumnekos LuaLS:

image 

image

Is there a way for me to adapt the format of my meta libraries to achieve a similar outcome to what I am used to?

### LINKED ISSUE #993: Feature request: ability to specify `--level` of output in `emmylua_check` command.
As for now, it will be great, if it were possible to disable the output of hints and warnings;
and leave only the output of errors.

_And maybe in the future you'll consider implementing something like [rule levels](https://phpstan.org/user-guide/rule-levels)_

### LINKED ISSUE #999: Wrong type resolution when inheriting array class
Hi, I have the following snippet:
```lua
---@class Foo
---@field [integer] string
---@field other number
local f = {}

---@param array string[]
local function bar(array)
  print(array[1])
end

---@param foo Foo?
local function baz(foo)
  foo = foo or { other = 5 } --[[@as Foo]]
  bar(foo)
end

baz(f)
```
This is fine so far, but dropping `--[[@as Foo]]` introduces a warning because `foo` is not resolved as `Foo` anymore and thus not as a valid `string[]`. Is this a design choice or would it be possible to have the right type resolution without the explicit comment? Thank yon in advance!

### LINKED ISSUE #1015: Snippet for pushing element to array
Hello! Another snippet that I used quite often with LuaLS completes `someTable[<CURSOR>]` into `someTable[#someTable + 1] = <CURSOR>`. I have not found it with emmylua, is it already implemented? How do I trigger it if it is? Thanks in advance!

### LINKED ISSUE #1019: Type inference ignores loops
This works:
```lua
---@class MyClass

local thing = nil
if true then
    thing = {} --[[@as MyClass]]
end

-- thing is now MyClass?
```

This doesn't:
```lua
---@class MyClass

local thing = nil
for i = 1, 10 do
    thing = {} --[[@as MyClass]]
end

-- thing is nil
```

happens for while as well:

```lua
---@class MyClass

local thing = nil
while true do
    thing = {} --[[@as MyClass]]
    break
end

-- thing is nil
```

### LINKED ISSUE #1032: Return type from assert breaks after asserting function further down in the file
It seems `assert()` breaks after asserting a method defined further down in the file. The returned type from `assert()` after that point becomes `()`.

```lua
--- @namespace assert_with_method_defined_later

--- @class Foo
local Foo = {}

--- @type string?
local nullable_string

--- @return boolean
function Foo:defined_before() return false end

function Foo:main()
  local _v1 = assert(nullable_string)
  --    ^ type: `string` ✅

  assert(self:defined_before())

  local _v2 = assert(nullable_string)
  --    ^ type: `string` ✅

  assert(self:defined_later())

  local _v3 = assert(nullable_string)
  --    ^ type: `()` ❌
end

--- @return boolean
function Foo:defined_later() return false end
```

Curiously it works if the returned value from the assert is assigned to a variable:

```lua
--- @namespace assert_with_method_defined_later

--- @class Foo
local Foo = {}

--- @type string?
local nullable_string

function Foo:main()
  local _ = assert(self:defined_later())

  local _v1 = assert(nullable_string)
  --    ^ type: `string` ✅
end

--- @return boolean
function Foo:defined_later() return false end
```

Reproduced on `0.22.0` and 2f9cb6d5.

### LINKED ISSUE #1112: hover return_overload
@lewis6991 
`@return_overload` Does it only take effect during flow? I think we should merge them one by one during hover.
For example, the following should be `local function parse() -> true|false, integer|string`
```rust
    #[gtest]
    fn test_return_overload_description_hover() -> Result<()> {
        let mut ws = ProviderVirtualWorkspace::new();
        check!(ws.check_hover(
            r#"
                ---@return_overload true, integer success
                ---@return_overload false, string failed
                local function parse()
                end

                local <??>alias = parse
            "#,
            VirtualHoverResult {
                value: "```lua\nlocal function parse()\n  -> true, integer\n  -> false, string\n\n```\n\n---\n\n@*return_overload* #1 — success\n\n@*return_overload* #2 — failed".to_string(),
            },
        ));
        Ok(())
    }
```

### LINKED ISSUE #1122: moduleMap does not appear to be applied correctly
In Factorio, our require paths can take the form of `require("__modname__/somefile.lua")` or `require("__modname__.somefile")`, for loading files from another mod. For most mod devs, stripping the `__` and `.lua` and looking for this in the ordinary fashion (in workspace or library paths, one of which will include the /mods tree) should be sufficient to make this actually resolve.

I have tried several variations of moduleMap rules
```json
"moduleMap": [
      {
        "pattern": "^__(.*)__(.*)$",
        "replace": "$1$2"
      },
      {
        "pattern": "^(.*)\\.lua$",
        "replace": "$1"
      }
    ]
```

```json
"moduleMap": [
      {
        "pattern": "__",
        "replace": ""
      },
    ]
```

but nothing seems able to make these able to resolve

emmylua understands these correctly, but they would not work in Factorio:
`local sigstr = require("signalstrings/signalstrings")`
`local sigstr = require("signalstrings.signalstrings")`

but not any of these, which are what would actually work in Factorio:
`local sigstr = require("__signalstrings__/signalstrings.lua")`
`local sigstr = require("__signalstrings__.signalstrings")`
`local sigstr = require("__signalstrings__/signalstrings")`

### LINKED ISSUE #1123: Incorrect type deduction around literal table with boolean keys
Image

The table
```lua
local control_wire = {
  [true] = defines.wire_connector_id.combinator_input_green,
  [false] = defines.wire_connector_id.combinator_input_red,
}
```
seems to misplace its keys in type deduction, as later on

```lua
local flip = not not self.flip_wires
local wire = control_wire[flip]
```
 gives `wire` a deduced type of `nil`

a literal `local wire = control_wire[true]` or `local wire = control_wire[false]` deduces correctly to the individual values

### LINKED ISSUE #1135: Alias definition before class weirdness with nilable type not functioning as optional field
I am taking a Factorio mod that was typed for LuaLS, and moving its types over to emmylua.
The first mod I am working on has a structure that may or may not contain a type that is itself recursive (with `nil` being a possible value).

I found that my tables were complaining about it missing the "required" field, and when investigating a simple case I found that whether or not the type has already been defined by the point of the class definition seems to determine the nilability behavior.

Instance 1:
```lua
---@alias B true?
---@class A
---@field field? B

---@type A
local var = {} -- Warning: Missing Required fields in type `A`: `field

---@type A
local var = {field = {}} -- No warning (wrong)

---@type A
local var = {field = "bleh"} -- Warning: Cannot assign `string` to `B`.
```

Instance 2:
```lua
---@class A
---@field field? B
---@alias B true?

---@type A
local var = {} -- No warning (correct)

---@type A
local var = {field = {}} -- Warning: Cannot assign `table` to `B?`

---@type A
local var = {field = "bleh"} -- Warning: Cannot assign `string` to `B?`
```

And more visible deeper insight (probably)
Instance 3:
```lua
---@alias B number|true
---@class A
---@field field? B

---@type A
local var = {field = {}} -- Warning: Cannot assign `table` to `(number|true)?`.
```

Instance 4:
```lua
---@class A
---@field field? B
---@alias B number|true

---@type A
local var = {field = {}} -- Warning: Cannot assign `table` to `B?`.
```

### LINKED ISSUE #1142: `keyof` doesn't play nice with other generic types
When `keyof` is in the return of a conditional type it seems to convert the keys into a tuple instead of returning them as a union
```lua
---@class A
---@field one 1
---@field two 2
---@field three 3

---@alias B<T> T extends any and keyof T or never
---@type B<A>
local tmp -- Type is `("one","two","three")`. Notice the `,`'s instead of `|`'s (I didn't for a while)
```
Constraints don't accept it when they should
```lua
---@alias C<K extends keyof A> any
-- type `keyof A` does not satisfy the constraint `("two"|"three"|"one")`. (despite the constraint being `keyof A`)
---@type C<keyof A>
local tmp
```

It seemingly resolves to `any` when used to index the type it's from instead of the expected union of values
```lua
---@alias KeyofA keyof A
---@type A[KeyofA]
local tmp = whatver_you_want -- *Nothing* will warn here
```

I'll also point out that EmmyLua does not support `K extends keyof T` in the type syntax. Whereas TypeScript does.
```lua
-- Complains about type here ∨
---@alias D<T,K> K extends keyof T and K or never
---@type D<A,'one'|'four'>
local tmp -- Type should be `"one"`
```
And I can't find any more places it's broken. Though I feel like I'm missing something

### LINKED ISSUE #1143: keyof type checking fails when used as a field type in ---@class
When using keyof to define a field type inside a ---@class, the type checking does not work as expected. The same keyof expression works correctly when used directly in a function parameter.

```lua
---@class c
---@field net table
---@field go table
local c = {}

---@class opt
---@field f keyof c  

---@param opt opt
function test(opt)
end

test({
    f = "net1"-- 这里传入 "net" 是不合法的，但类型检查没有生效
})

---@param f keyof c
function test2(f)
end

-- 这里传入 "net1" 会触发类型错误，类型检查正常工作
test2("net1")

```

xuhuanzy:
1. 重构泛型, 见 #1096
2. 移除 `Attribute` 标签, 现在他的定义行为与 c# 的类似
3. 重构 hover signature
4. `emmylua_check` 添加选项`severity`以过滤输出
5. 一堆 issue 修复

fixes #999
fixes #912
fixes #560
fixes #903
fixes #860
fixes #632
fixes #626
fixes #711
fixes #1123
fixes #1122
fixes #1135
fixes #1143
fixes #1142


xuhuanzy:
@lewis6991 Can you take a look at the implementation of c5d947d120d3447c23d756de765660b58b22a6ac? Does it have a significant impact on performance?

lewis6991:
I hope a prompted AI review suffices:

---

  - Correctness: `branch_flow_has_direct_never_call` only works when `branch_flow_id` is a TrueCondition/FalseCondition node. But merge inputs are the last flow node in the branch. If the
    branch has a local, assignment, cast, etc. before `panic()`, the helper returns None and keeps the branch. Example still likely fails:

    ```lua
    if num1 == nil then
        local reason = "bad"
        panic()
    end
    accept_num(num1)
    ```

    Relevant lines: c5d947d:crates/emmylua_code_analysis/src/semantic/infer/narrow/get_type_at_flow.rs:1875-1892.

  - Perf: yes, some concern. `evaluate_walk` now filters every `BranchLabel` query, allocates a `Vec`, scans branch AST, and may call `try_infer_expr_no_flow` for direct calls. No-flow caches soften this, but branch merges are hot and this filtered reachability result is not cached. Relevant lines: .../get_type_at_flow.rs:1380-1389, 1858-1872, 1923-1933.

  - Type-system coherence: The desired behavior is right, but this should be modeled as whether a branch can continue running, not as a special never filter during type merging.

    `LuaType::Never` means “this expression does not produce a value.”
    The branch question is different: “can execution get past this branch?”

    For example:

    ```lua
    if num1 == nil then
      panic() -- returns never
    end

    accept_num(num1)
    ```

    After `panic()`, execution cannot continue to `accept_num`. So the `num1 == nil` branch should not contribute any type for `num1` after the if. Only the branch where `num1 ~= nil` reaches that point, so `num1` should be integer.

    The concern with the commit is that it teaches the type-merging code to scan branch bodies for direct never calls and drop those branches. That mixes two separate ideas: expression types and whether execution can continue. The cleaner model is for the flow analysis to record that a `---@return never` call stops execution at that point, so the existing branch merge logic only sees branches that can actually continue.


  I’d request changes. The better shape is to model “this branch falls through” directly, or at least walk from the branch merge input back to its branch condition before deciding whether the corresponding block is terminated.

xuhuanzy:
> I’d request changes. The better shape is to model “this branch falls through” directly, or at least walk from the branch merge input back to its branch condition before deciding whether the corresponding block is terminated.

My original intention was to make minimal changes to the existing flow analysis, because `never` was previously equivalent to `nil`. Therefore, at this stage, a conservative analysis must be performed.

xuhuanzy:
@lewis6991 can it be confirmed that `flow` did not have special handling for `never` in the previous implementation?

lewis6991:
Yeah, I think I overstated this a bit.

Current `main` does already give `never` some special meaning in flow. It is used in condition narrowing / branch merging for impossible paths, so saying "flow had no special handling for `never`" is not really accurate.

What I had in mind was the cleanup in #1098, where unreachable branches are separated from real `never` values. Since that PR is already open, I would prefer to merge #1098 first and then rebase this change on top of it. That should let this fix use the cleaner flow model instead of adding more logic around the current `LuaType::Never` behavior.

The user-visible behavior here is still right: a `---@return never` call should narrow the same way as `error()`.

After #1098, I think this would ideally be modeled closer to the existing `error()` handling: when a call statement is known to return `never`, flow should record that execution stops at that call. Then the branch merge naturally only sees paths that can continue, instead of scanning branch bodies during type merging.

Roughly:

```rust
} else if call_expr.is_error() || call_expr_returns_never(...) {
    if let Some(ast) = LuaAst::cast(call_expr.syntax().clone()) {
        bind_each_child(binder, ast, current);
    }

    let return_flow_id = binder.create_return();
    binder.add_antecedent(return_flow_id, current);
    return_flow_id
} else {
    if let Some(ast) = LuaAst::cast(call_expr.syntax().clone()) {
        bind_each_child(binder, ast, current);
    }

    current
}
```

xuhuanzy:
@lewis6991 So, can you resolve #877 together with #1098? That way I can remove the commit c5d947d120d3447c23d756de765660b58b22a6ac.

xuhuanzy:
The handling of #877 has been removed, pending the merge of #1098.

PR #1130, opened 2026-06-25
fix: preserve explicit doc types for self members and hover

wang573287909:
## Problem

`---@type` annotations on `self.member = nil` / `self.member = {}` inside method bodies were not preserved:
1. **Inference**: The doc type was not registered on the class owner, so `controller.mInviterHead` inferred as `nil` instead of `p_role_head`.
2. **Hover**: The hover displayed the RHS inferred type (`nil` / `table`) instead of the explicit doc type (`p_role_head` / `ThunderStrikeGuardianDataStruct[]`).

## Fix

1. **stats.rs**: For `LuaType::Ref` prefix in `set_index_expr_owner`, only register the member in the class owner's member map when the prefix is implicit `self` (not arbitrary `---@type Foo` variables).
2. **build_hover.rs**: When building member hover, check if the member has an explicit doc type (`type_cache.is_doc()`), and prefer it over the RHS-inferred hover type.
3. **Tests**: Added regression tests for both inference and hover.

## Test plan

- [x] `cargo test --workspace` (via CI)
- [ ] Verify `self.mInviterHead = nil` with `---@type p_role_head` infers correctly from instance access
- [ ] Verify hover shows doc type for nil and table initializer assignments
- [ ] Verify `Foo.extra = 1` does NOT pollute other `---@type Foo` instances

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,6 +6,7 @@ on:
     tags: ['v*']
   pull_request:
     branches: [main, master]
+  workflow_dispatch:
 
 jobs:
   build:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,134 +1,70 @@
-name: Rust
+name: Build emmylua_ls
 
 on:
   push:
-    branches:
-     - main
-    tags:
-     - "*"
+    branches: [main, master]
+    tags: ['v*']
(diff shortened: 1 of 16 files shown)


CppCXY:
You changed a bunch of content that is unrelated to the PR. Please reorganize the PR before considering submission.

Issue #1178, opened 2026-07-16
Overload with mismatched types fails on self-assign


CodeGreen0386:
Minimal reproduction:
```
---@class Vector
---@operator div(number): Vector

---@return Vector
local function vector() return {} end

---@return number
local function number() return 1 end

local v = vector()

local position = vector()
position = position / number()

v = position
```

Should get a "Cannot assign `number` to `Vector`. " error on `v`.

Prommah:
~I think the problem is actually just an issue with operators in general~. Edit: Nope, seems I'm completely wrong and these are unrelated. Sorry. Leaving the rest of the comment here since it's still an issue I observed when looking into this.

Here's an example straight from [the operator docs](https://github.com/EmmyLuaLs/emmylua-analyzer-rust/blob/98c85ace2375632ebfdb0cf3ccdb7703773c70f8/docs/emmylua_doc/annotations_EN/operator.md) which I've trimmed down:

```
---@class Vector
---@field x number
---@field y number
---@operator add(Vector): Vector
local Vector = {}

function Vector.new(x, y)
    return setmetatable({x = x or 0, y = y or 0}, {
        __index = Vector,
        __add = Vector.__add
    })
end

function Vector:__add(other)
    return Vector.new(self.x + other.x, self.y + other.y)
end

local v1 = Vector.new(1, 2)
local v2 = Vector.new(3, 4)

local v3 = v1 + v2
local test = v3.x
```

At the end, `v3` is incorrectly `number` and accessing `v3.x` gives ``Undefined field `x`.``

MuMieShi:
We hit the same class of issue in a production static-analysis pipeline (game-engine Lua with heavily overloaded vector math). One extra data point that may help localize it: with self-assignment, whether the type survives depends on **which overload signature gets used**.

Self-contained repro (no external defs):

```lua
---@class Vec
---@field x number
---@operator add(number): Vec
---@operator add(Vec): Vec
---@operator mul(number): Vec

---@return Vec
local function vec() return { x = 0 } end

---@param s number
local function f(s)
    local pos = vec()

    local r = pos * s          -- fresh local: r inferred as Vec
    local rx = r.x             -- OK

    pos = pos + vec()          -- self-assign, Vec-param overload
    local px1 = pos.x          -- OK

    pos = pos * s              -- self-assign, number-param overload
    local px2 = pos.x          -- ERROR: Undefined field `x`
end
```

[...]

PR #1182, opened 2026-07-16
fix(flow): preserve self-assignment antecedent types

### LINKED ISSUE #1180: Flow analysis breaks the integer property of the index after assignment.
```lua
function test()
    local arr = {1, 2, 3, 4, "hello", 6, false, 7}

    local index = 1

    while index <= #arr do
        local v = arr[index]

        if type(v) == "boolean" then
            table.remove(arr, index)
            index = index - 1
            print(index) -- number
        end

        index = index + 1 -- number
    end
end
```

@lewis6991

lewis6991:
## Problem

Self-dependent assignments replaced the RHS read with `unknown` to avoid
recursive replay. For expressions such as `index = index - 1`, this discarded
the pre-assignment `integer` type and widened the result to `number`.

## Solution

- resolve unavoidable self-reads through the iterative flow scheduler
- distinguish variable versions by both `VarRefId` and `FlowId`
- preserve linear repeated-assignment replay without bypassing skippable index
  dependencies
- remove the `unknown` injection and its `and`/`or` exception

## Tests

- `cargo test -p emmylua_code_analysis issue_1180`
- `cargo test -p emmylua_code_analysis test_issue_1114_repeated_self_dependent_assignments_build_semantic_model`
- `cargo test -p emmylua_code_analysis test_issue_1116_generic_call_index_replay_builds_semantic_model`
- `cargo test -p emmylua_code_analysis` (1,079 tests)
- `cargo fmt --all --check`
- `git diff upstream/main...HEAD --check`

Fixes #1180

### DIFF
--- a/crates/emmylua_code_analysis/src/compilation/test/flow.rs
+++ b/crates/emmylua_code_analysis/src/compilation/test/flow.rs
@@ -3600,6 +3600,48 @@ _2 = a[1]
     }
 
     #[test]
+    fn test_issue_1180_self_assignment_replays_antecedent_type() {
+        let mut ws = VirtualWorkspace::new();
+        ws.def(
+            r#"
+        local index = 1
+        index = index - 1
+        after_assign = index
+        "#,
+        );
+
+        assert_eq!(ws.expr_ty("after_assign"), ws.ty("integer"));
+    }
+
+    #[test]
+    fn test_issue_1180_while_loop_preserves_integer_self_assignments() {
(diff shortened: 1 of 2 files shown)


xuhuanzy:
I think you can also take a look at this #1178
This issue is related to it.

lewis6991:
Thanks, I checked #1178. This PR fixes the original example: position remains a Vector after the self-assignment, so `v = position` should not report an error.

The other example in the comments still fails, but it does not involve self-assignment and looks like a separate custom-operator issue. I think that should be handled separately.

PR #1209, opened 2026-08-07
fix(docs): wrap and indent overflowing signatures

Nerixyz:
Long signatures would cut off, and you'd need to scroll/select the text to see everything. With this change, the text wraps. The wrapped text is indented.

Before: 

After:


Nerixyz:
It would be nice if the JS and CSS used a formatter and a linter. You might want to check https://biomejs.dev/ which is also written in Rust.

PR #1228, opened 2026-08-19
fix(type-check): accept exhaustive boolean unions

### LINKED ISSUE #1226: `true | false` is assignalbe to `boolean`, but `boolean` isn't assignable to `true | false`
```lua
---@return boolean
local function test1()
  return true ---@as true | false -- OK
end

---@return true | false
local function test2()
  return true ---@as boolean -- Annotations specify that return value 1 has a type of `(true|false)`, returning value of type `boolean` here instead. 
end
```

lewis6991:
## Problem

`boolean` values were rejected where the equivalent `true | false` literal
union was expected.

## Solution

- Treat unions containing both boolean literals as exhaustive during type
  checking.
- Reuse structural union construction in both document type inference paths.

## Tests

- `cargo test -p emmylua_code_analysis`
- `cargo fmt --all -- --check`
- `cargo clippy -p emmylua_code_analysis --all-targets --all-features --locked -- -D warnings`

Fixes #1226

### DIFF
--- a/crates/emmylua_code_analysis/src/compilation/analyzer/doc/infer_type.rs
+++ b/crates/emmylua_code_analysis/src/compilation/analyzer/doc/infer_type.rs
@@ -511,27 +511,9 @@ fn infer_binary_type(
         let right_type = infer_type(analyzer, right);
         if let Some(op) = binary_type.get_op_token() {
             match op.get_op() {
-                LuaTypeBinaryOperator::Union => match (left_type, right_type) {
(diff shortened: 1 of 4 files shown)


xuhuanzy:
I’m refactoring the type-checking system, and it’s nearly complete. I think it would be better for me to handle this issue.

lewis6991:
Doesn't it matter? AI can trivially resolve any conflicts.

PR #1229, opened 2026-08-20
fix(generic): preserve variadic return tails

### LINKED ISSUE #1227: Generic Variadic returns don't work combined with any other return defined
```lua
---@generic T
---@param t T
---@return int, T...
local function some_func(t) end

local tmp1, tmp2, tmp3 = some_func("") -- tmp3 is unknown
```
I found this when trying to define the return of some function as `T, T...` since it made it complain if not at the end of an array definition (which might be a bug on its own, but I'd need a feature to replace it...)

In my playing with the example, I found that just swapping out the generic `T` with a static type like `string` it suddenly can understand all the returns.

I also found that `return_overload` seems to break slightly differently
```lua
---@generic T
---@param t T
---@return_overload int, T...
local function some_func(t) end

local tmp1, tmp2, tmp3, tmp4 = some_func("") -- tmp4 is unknown
```
No clue why  `return_overload` specifically treats a variadic return like a "and one more"

To be clear, *this* (and with `return_overload`) works correctly:
```lua
---@generic T
---@param t T
---@return T...
local function some_func(t) end

local tmp1, tmp2, tmp3 = some_func("") -- all are strings
```

lewis6991:
## Problem

`T...` variadic return tails were unwrapped to one `T` when they followed
fixed return values. Deeper assignment slots therefore became unknown,
including for `@return_overload`.

## Solution

- Keep instantiated unbounded base variadics wrapped in multi-return rows.
- Continue expanding finite multi-return substitutions.

## Tests

- `cargo test -p emmylua_code_analysis`
- `cargo fmt --all --check`
- `cargo clippy -p emmylua_code_analysis --all-targets --all-features --locked -- -D warnings`

Fixes #1227

### DIFF
--- a/crates/emmylua_code_analysis/src/compilation/test/generic_test.rs
+++ b/crates/emmylua_code_analysis/src/compilation/test/generic_test.rs
@@ -1672,6 +1672,24 @@ mod test {
     }
 
     #[test]
+    fn test_generic_variadic_return_after_fixed_return_keeps_deep_slots() {
+        let mut ws = VirtualWorkspace::new();
+        ws.def(
+            r#"
+            ---@generic T
+            ---@param value T
+            ---@return integer, T...
+            local function some_func(value)
+            end
+
+            First, Second, Third = some_func("")
+            "#,
+        );
+
+        assert_eq!(ws.expr_ty("Third"), ws.ty("string"));
+    }
+
+    #[test]
     fn test_nested_function_keeps_unresolved_variadic_return_generic() {
         let mut ws = VirtualWorkspace::new();
         ws.def(
--- a/crates/emmylua_code_analysis/src/compilation/test/return_overload_generic_test.rs
+++ b/crates/emmylua_code_analysis/src/compilation/test/return_overload_generic_test.rs
@@ -61,6 +61,24 @@ mod test {
     }
 
     #[test]
+    fn test_generic_variadic_return_overload_after_fixed_return_keeps_deep_slots() {
+        let mut ws = VirtualWorkspace::new();
+        ws.def(
+            r#"
+            ---@generic T
+            ---@param value T
+            ---@return_overload integer, T...
+            local function some_func(value)
+            end
+
+            First, Second, Third, Fourth = some_func("")
+            "#,
+        );
+
+        assert_eq!(ws.expr_ty("Fourth"), ws.ty("string"));
+    }
+
+    #[test]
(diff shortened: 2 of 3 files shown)


CppCXY:
I am about to merge the salsa3 branch within the next one to two weeks, so I do not plan to merge anything before that, as there will be a lot of conflicts.

However, if things do not proceed as expected, I will merge this PR and defer the salsa update.

lewis6991:
@CppCXY any update on merging the salsa branch? I've been holding off raising any more PR's.

CppCXY:
I've been busy with work for the past two weeks, so I haven't updated much. I've given up referencing the current salsa library; because the implementation's performance is too poor, I've done a lot of rollbacks. However, the current development branch is still salsa3. It currently lacks some core generic support and proper support for flow, so it isn't usable yet. Also, I've done extensive and wide-ranging redesigns, and none of the current PRs are mergeable at the moment, so we'll need to wait longer. And since China's National Day holiday is coming soon, the wait may be even longer.

CppCXY:
In addition, while exploring related technologies, I picked up a project I created a year ago and plan to give it a try., feel free to take a look: https://github.com/CppCXY/cpp_ls

lewis6991:
Would it make sense if I raise PR's to the salsa branch or is it too rough still?

CppCXY:
I'm not opposed to it, and we can give it a try. The difficulty, though, is that I deleted most of the previous code—around 80,000 lines of changes.

lewis6991:
How is the testsuite? Did you remove tests too? I don't mind about what the changes are as long as the behavior is correct and reinforced with tests.

xuhuanzy:
This process may be quite complex. I’m also maintaining a refactor of a type checker spanning more than 10,000 lines, involving caching and instantiation. I don’t think this is the right time to submit a PR related to generics.

lewis6991:
It's been over a month. Is progress on this project really grinded to a halt?

Instead of maintaining all these huge rewrites, why not keep the main branch active and use AI keep these rewrite branches rebased? That way we can keep collecting test cases for important behaviours we need to implement.

CppCXY:
The main problem here is that your PRs are almost entirely working on content that I deleted. However, if you want to get involved in the development of this branch's content right now, I can plan to push the merge forward next Monday. Even though there are many bugs in the functionality, it's still better than appearing inactive.

PR #1254, opened 2026-09-21
feat(check): add GitLab Code Quality output

yeshan333:
## Summary

Add native GitLab Code Quality output to `emmylua_check`.

## Changes

- Add `-f gitlab` with stdout and file output.
- Emit repository-relative findings with GitLab severities and deterministic fingerprints.
- Document GitLab CI usage.
- Keep the workspace clean under the current Clippy release.

## Verification

- `cargo test -p emmylua_check --all-features` — 6 passed
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- End-to-end fixtures — 3 findings for invalid code, `[]` for clean code


yeshan333:
Addressed the actionable review points in `a23fe210`:

- canonicalize both the project root and analyzed file before computing the report path;
- reject files outside the repository root with a clear warning instead of emitting an unusable path;
- add a regression test for the outside-root case;
- make report creation, serialization, and write failures include actionable context;
- document why the repository root prefers `CI_PROJECT_DIR`/CWD over the analyzed workspace (the workspace may be a subdirectory such as `src`);
- make the `--output` help text explicitly name `json`, `sarif`, and `gitlab`;
- document the severity mapping rationale.

The start line remains part of the fingerprint intentionally. Without a location component, two identical diagnostics with the same rule and message in one file collapse to one GitLab finding. It also keeps native output compatible with the existing downstream SARIF converter. The documentation now calls these fingerprints deterministic and explicitly states that moving a finding changes its identity.

Changing the `OutputWriter` trait to return `Result` would be a wider refactor across every existing writer. This PR follows the current trait while replacing bare I/O/serialization `unwrap()` calls in the new writer with contextual failure messages.

yeshan333:
Follow-up in the latest commit: I removed the downstream-converter compatibility coupling and simplified repository-path handling.

- The current working directory is now the single repository-root convention; run the command from the repository root.
- Fingerprints are defined solely by the native output contract: rule, repository-relative path, full diagnostic range, and description.
- The `CI_PROJECT_DIR` fallback chain and extra path-normalization layers were removed.
- Files outside the repository root are still skipped because GitLab Code Quality requires repository-relative paths.

The PR description and tests have been updated to match this behavior.

yeshan333:
The latest follow-up keeps the simplified design while fixing cross-platform path correctness:

- reuse the analyzer's existing `normalize_local_path` function so the working directory and VFS paths use the same representation on Windows;
- normalize each analyzed file before checking that it belongs to the repository, which also resolves `..` and symlinked paths;
- build report paths from `Path` components instead of replacing every backslash, preserving valid backslashes in Unix filenames;
- replace the conversion-only assertion with a full `write` → `finish` report test, and align the remaining test names/assertions with the documented contract.

No downstream-converter compatibility behavior was added back.

yeshan333:
CI fix pushed in `a899275b`. The updated branch passes Clippy 1.98 locally; GitHub now requires approval again before the fork workflows can run.

CppCXY:
It’s a good addition. Although I don’t use GitLab, since it involves some things I’m bound to modify, I don’t plan to merge it right now—I’ll wait until I’ve finished my refactoring.

